PR #38, opened 2014-07-03
Fixed emscripten build

UplinkCoder:
Hello I just fixed a few header-files blocking the build with emcc.
However, there is still an error that requires commenting out the `_LARGEFILE_64` defines in `emscripten/system/include/libc/stdio.h` I have not yet found a neat fix for that.
And one still needs `llvm-tblgen`, `clang-tblgen` and `llvm-config` form a _native build_

### DIFF
--- a/include/llvm/Support/Host.h
+++ b/include/llvm/Support/Host.h
@@ -16,7 +16,7 @@
 
 #include "llvm/ADT/StringMap.h"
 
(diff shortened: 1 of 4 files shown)


sbc100:
This repository is no longer active - Emscripten is now based
on the upstream llvm WebAssembly backend.

PR #48, opened 2014-09-08
Support added for exceptions with ASYNCIFY

NWilson:
_See matching emscripten pull request [#2772](https://github.com/kripken/emscripten/pull/2772)._

We add support for exceptions with asynchronous functions. Changes:
1. The LowerEmExceptions pass turns "invoke" instructions into "call" instructions plus some gloop. Rather than trying to recognise the gloop, I've added a function attribute "emscripten_invoke" to the resulting call instruction so that LowerEmAsyncify can easily spot which call instructions are inside try blocks.
2. LowerEmAsyncify mangles the invoke sequence by splitting the block right after the call instruction. I've fixed this up by adding a call to "emscripten_async_postinvoke1" just before the split on the other side of the split "emscripten_async_postinvoke2". These functions do nothing, they just set and assert the `InvokeState` variable so that illegal reordering of the invoke sequence by optimisations will be detected.
3. Finally, we've added a third instruction to `emscripten_alloc_async_context` so that a flag is set in the context indicating where it's a checked asynchronous call (invoke) or unchecked.

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -158,16 +158,6 @@ DEF_CALL_HANDLER(emscripten_landingpad, {
 DEF_CALL_HANDLER(emscripten_resume, {
   return "___resumeException(" + getValueAsCastStr(CI->getOperand(0)) + ")";
 })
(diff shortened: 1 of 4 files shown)


NWilson:
Closing ancient PR

PR #59, opened 2015-01-08
fix for userId's > 6 digits

leemorgan-mobile:


tools/llvm-ar/llvm-ar.cpp
```diff
@@ -579,11 +579,24 @@ computeNewArchiveMembers(ArchiveOperation Operation,
...
-  OS << Data;
-  unsigned SizeSoFar = OS.tell() - OldPos;
-  assert(Size >= SizeSoFar && "Data doesn't fit in Size");
-  unsigned Remaining = Size - SizeSoFar;
+  std::string buffer;
+  raw_string_ostream rss(buffer);
+  rss << Data;
+  bool cropped = rss.str().size() > Size; // use rss.str() which flushes
+  unsigned Remaining = 0;
+  if( cropped )
```
sunfishcode: Please follow the local coding style for whitespace and braces (space after the 'if', no space after '(', no space before ')', '{' on the same line as the 'if'/'else').

tools/llvm-ar/llvm-ar.cpp
```diff
@@ -600,8 +613,8 @@ static void printRestOfMemberHeader(raw_fd_ostream &Out,
                                     unsigned GID, unsigned Perms,
                                     unsigned Size) {
   printWithSpacePadding(Out, ModTime.toEpochTime(), 12);
-  printWithSpacePadding(Out, UID, 6);
-  printWithSpacePadding(Out, GID, 6);
+  printWithSpacePadding(Out, (UID > 999999u ? 999999u: UID), 6);
```
sunfishcode: It seems like saturating the UID and GID is likely to produce subtly and silently incorrect results whenever it activates. Unless I'm misunderstanding something here, issuing an error would be more appropriate, to let the user know that it isn't possible for llvm-ar to perform what it has been asked to perform.

sunfishcode:
Also, ideally this patch should be submitted upstream to LLVM, as it looks like it would affect all LLVM users, and so that the Emscripten maintainers don't have to maintain it as an out-of-tree patch. You can file a bug at http://llvm.org/bugs/ . Thanks!

tools/llvm-ar/llvm-ar.cpp
leemorgan-mobile: I agree, any number is wrong in this instance. But I don't want the build
to break just because .a files only hold 6 digits. My initial instinct was
to have it blank but I don't know what other tools I'll break that expect a
number. Changing the .a format is the real fix!

Cheers.

sunfishcode: I agree that breaking the build isn't nice, but silently doing something other than what was requested is even less nice. What do other ar implementations do in this situation?

leemorgan-mobile: Well what was requested was a build - I'm not even sure why that
information is in the file.  I can't imagine other implementations break
the build so perhaps then the field is made blank. I'll check on my machine
tomorrow.

kripken:
I agree, unless there is an urgent reason to fix this in emscripten, we should focus on fixing it upstream in LLVM. Closing this, but correct me if I'm wrong and there is an urgency here that we need to address, and we'll reopen.

kripken:
Ah, I didn't connect this to the mailing list post. So the problem this addresses is quite severe.

Do we know why it's so rare? And is there a reason for it to hit emscripten more often than anything else? I suppose it's just that we are big users of llvm-ar?

Reopening for now.

leemorgan-mobile:
No. It's a surprise to be hitting it now. From what I could ascertain
userid used to be limited and within 6 digits in Linux but hasn't for a
while. Also distributions may have had their own limits. I'm on ubuntu but
in corporate windows exchange network which may play a part in my high Id.

Still it doesn't feel right that it hasn't been fixed already given llvms
use. I half presumed emscriptens llvm came from an old version.

I'll look into submitting upstream tomorrow which may provide some answers.

leemorgan-mobile:
So it appears this was fixed a year ago.

Although their solution is to seek back leaving the higher 6 digits of any userID. This seems the worst of all solutions to me!

gcc etc appear to set the ID's to zero, which is the root ID.

Anyway - when do you think emscripten's repo will will catch up?

leemorgan-mobile:
Actually I think the fix is still buggy. They write the digits, then seek without knowing if a flush has happened. It probably needs another patch.

kripken:
The incoming branch contains LLVM 3.5, as of a few days ago, which is less than a year old - so their fix should be on that.

sunfishcode:
FTR, the upstream patch is r199223.

kripken:
Hmm, how can I verify that that commit is in our 3.5 copy? I don't see anywhere I can browse llvm svn by revisions.

leemorgan-mobile:
Ah ok, the write/flush isn't an issue as the seek flushes first and seeks on the file descriptor.

I can see the fix in incoming but not in latest.

sunfishcode:
I don't know of an automatic way, but I've now manually confirmed that the patch is in our 3.5 copy.

I looked into this issue a little more, and it now seems like the UID field is completely meaningless, so it isn't actually important what goes there. Always using 0 sounds like the best fix, as it would eliminate differences between otherwise identical builds, but it's not super important.

Issue #60, opened 2015-01-09
Flatten global errors on incoming


trippleflux:
I am using visual studio 2013 to compile latest incoming git, but stuck at this following : 
```
emscripten-fastcomp\lib\Transforms\NaCl\FlattenGlobals.cpp(178): error C2039: 'getRelocUse' : is not a member of '`global namespace''
emscripten-fastcomp\lib\Transforms\NaCl\FlattenGlobals.cpp(178): error C2660: '`anonymous-namespace'::FlattenedConstant::Reloc::getRelocUse' : function does not take 1 arguments
```

kripken:
Must be a problem after the LLVM 3.5 update.

Odd that the code builds in clang and gcc, but not in msvc. 

Does this patch fix it for you?

````
diff --git a/lib/Transforms/NaCl/FlattenGlobals.cpp b/lib/Transforms/NaCl/FlattenGlobals.cpp
index c1ba780..bf74e58 100644
--- a/lib/Transforms/NaCl/FlattenGlobals.cpp
+++ b/lib/Transforms/NaCl/FlattenGlobals.cpp
@@ -82,7 +82,7 @@ namespace {
   typedef std::vector<FlattenedGlobal*> FlattenedGlobalsVectorType;
 
   // Returns the corresponding relocation, for the given user handle.
-  static Constant *getRelocUse(RelocUserType *RelocUser) {
+  static Constant *_getRelocUse(RelocUserType *RelocUser) {
     return cast<Constant>(RelocUser->getReturnValue());
   }
 
@@ -175,7 +175,7 @@ namespace {
    public:
 
       unsigned getRelOffset() const { return RelOffset; }
-      Constant *getRelocUse() const { return ::getRelocUse(RelocUser); }
+      Constant *getRelocUse() const { return _getRelocUse(RelocUser); }
       Reloc(FlattenGlobalsState &State, unsigned RelOffset, Constant *NewVal)
           : RelOffset(RelOffset),
             RelocUser(State.getRelocUserHandle(NewVal)) {}
@@ -512,7 +512,7 @@ void FlattenGlobalsState::removeDeadInitializerConstants() {
   // Do cleanup of old initializers.
   for (RelocMapType::iterator I = RelocMap.begin(), E = RelocMap.end();
        I != E; ++I) {
-    getRelocUse(I->second)->removeDeadConstantUsers();
+    _getRelocUse(I->second)->removeDeadConstantUsers();
   }
 
 }
````

[...]

PR #61, opened 2015-01-12
VS2013 fix for getRelocData.

juj:
Work around Visual Studio bug https://connect.microsoft.com/VisualStudio/feedback/details/1085387 .


juj:
This fixes #60. While it's also possible to fix this in other ways, I thought to make the workaround as explicit as possible.

kripken:
@jfbastien - pnacl probably would benefit from this too.

jfbastien:
@juj: would making this non-`static` fix the issue? It's redundant in an anonymous namespace.

jfbastien:
I discussed this with clb on IRC, here's a different fix:
  https://codereview.chromium.org/828553007

PR #64, opened 2015-03-07
Optimize stack restoration.

waywardmonkeys:
Only emit stack restoration when the stack has been bumped by the
current function. This means that all functions must exercise
proper stack discipline.

This is done by looking for any static allocas during the computation
of the frame offset as well as any dynamic allocas when emitting the
code for the function.

This addresses https://github.com/kripken/emscripten/issues/3071.

cc: @chadaustin

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -155,6 +155,7 @@ namespace {
     int InvokeState; // cycles between 0, 1 after preInvoke, 2 after call, 0 again after postInvoke. hackish, no argument there.
     CodeGenOpt::Level OptLevel;
     const DataLayout *DL;
+    bool StackBumped;
 
     #include "CallHandlers.h"
 
@@ -162,7 +163,7 @@ namespace {
     static char ID;
     JSWriter(formatted_raw_ostream &o, CodeGenOpt::Level OptLevel)
       : ModulePass(ID), Out(o), UniqueNum(0), NextFunctionIndex(0), CantValidate(""), UsesSIMD(false), InvokeState(0),
-        OptLevel(OptLevel) {}
+        OptLevel(OptLevel), StackBumped(false) {}
 
     virtual const char *getPassName() const { return "JavaScript backend"; }
 
@@ -1746,7 +1747,9 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
   case Instruction::Ret: {
     const ReturnInst* ret =  cast<ReturnInst>(I);
     const Value *RV = ret->getReturnValue();
-    Code << "STACKTOP = sp;";
+    if (StackBumped) {
+      Code << "STACKTOP = sp;";
+    }
     Code << "return";
     if (RV != NULL) {
       Code << " " << getValueAsCastParenStr(RV, ASM_NONSPECIFIC | ASM_MUST_CAST);
@@ -1919,6 +1922,13 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
   }
   case Instruction::Alloca: {
     const AllocaInst* AI = cast<AllocaInst>(I);
+
+    // We've done an alloca, so we'll have bumped the stack and will
+    // need to restore it.
+    // Yes, we shouldn't have to bump it for nativized vars, however
+    // they are included in the frame offset, so the restore is still
+    // needed until that is fixed.
+    StackBumped = true;
 
     if (NativizedVars.count(AI)) {
       // nativized stack variable, we just need a 'var' definition
@@ -2427,6 +2437,7 @@ void JSWriter::printFunction(const Function *F) {
   nl(Out);
 
   Allocas.clear();
+  StackBumped = false;
 }
 
 void JSWriter::printModuleBody() {


kripken:
Please run the main test suite and other on this, `tests/runner.py` and `tests/runner.py other`.

waywardmonkeys:
Pretty sure that ``tests/runner.py`` was fine the last time I ran it ... I had to fix my patches a couple of times to get them working.

I just ran ``tests/runner.py other`` and it got these failures:

````
======================================================================
ERROR: test_crunch (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 2301, in test_crunch
    assert os.stat('test.data').st_size < 0.25*os.stat('ship.dds').st_size, 'Compressed should be much smaller than dds'
OSError: [Errno 2] No such file or directory: 'test.data'

======================================================================
ERROR: test_llvm_nativizer (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 2188, in test_llvm_nativizer
    output = Popen([os.path.join(self.get_dir(), 'files.o.run')], stdin=open(os.path.join(self.get_dir(), 'stdin')), stdout=PIPE, stderr=PIPE).communicate()
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

======================================================================
ERROR: test_scons (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 2132, in test_scons
    Popen(['scons']).communicate()
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory

======================================================================
FAIL: test_llvm_lit (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 2911, in test_llvm_lit
    assert p.returncode == 0, 'LLVM tests must pass with exit code 0'
AssertionError: LLVM tests must pass with exit code 0

======================================================================
FAIL: test_outline (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 989, in test_outline
    args=['-I' + path_from_root('tests', 'zlib')], suffix='c')
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 951, in test
    assert len(JS_ENGINES) > 1
AssertionError

======================================================================
FAIL: test_simplify_ifs (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 3031, in test_simplify_ifs
    ''', [3, 1, 1])
  File "/Users/bruce/Development/emscripten/tests/test_other.py", line 2972, in test
    assert ifs == actual_ifs, main + ' : ' + str([ifs, actual_ifs])
AssertionError: function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 if (_getenv(8) | 0) if (_getenv(16) | 0) {
  _puts(56) | 0;
  _puts(40) | 0;
  return 0;
 }
 _puts(24) | 0;
 _puts(40) | 0;
 return 0; : [3, 2]

----------------------------------------------------------------------
Ran 138 tests in 992.093s
````

I'm not sure which of those happen on incoming vs don't ...

**Comment by waywardmonkeys** [comment]
Looks like the simplify_ifs failure is new ... I will look into that tomorrow. (2am here)

**Comment by waywardmonkeys** [comment]
Okay, prior to my change, this code:

````c++
      #include <stdio.h>
      #include <stdlib.h>

      int main(int argc, char *argv[]) {
        if (getenv("A") && getenv("B")) {
            printf("hello world\n");
        } else {
            printf("goodnight moon\n");
        }
        printf("and that's that\n");
        return 0;
      }
````

was getting compiled in the test case to:

````javascript
['-g2'] 3
3 == 3
function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 var i3 = 0;
 i2 = STACKTOP;
 if (!(_getenv(8) | 0)) i3 = 4; else if (!(_getenv(16) | 0)) i3 = 4; else _puts(56) | 0;
 if ((i3 | 0) == 4) _puts(24) | 0;
 _puts(40) | 0;
 STACKTOP = i2;
 return 0;
````

But now, it is compiled to:

````javascript
function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 if (_getenv(8) | 0) if (_getenv(16) | 0) {
  _puts(56) | 0;
  _puts(40) | 0;
  return 0;
 }
 _puts(24) | 0;
 _puts(40) | 0;
 return 0;
````

**Comment by waywardmonkeys** [comment]
For the record, here are all of the compilation variations on that test source:

````
['-g2'] 3
2 == 3
function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 if (_getenv(8) | 0) if (_getenv(16) | 0) {
  _puts(56) | 0;
  _puts(40) | 0;
  return 0;
 }
 _puts(24) | 0;
 _puts(40) | 0;
 return 0;
['-profiling'] 1
1 == 1
function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 if ((_getenv(8) | 0) != 0 ? (_getenv(16) | 0) != 0 : 0) {
  _puts(56) | 0;
  _puts(40) | 0;
  return 0;
 }
 _puts(24) | 0;
 _puts(40) | 0;
 return 0;
['--profiling'] 1
1 == 1
function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 if ((_getenv(8) | 0) != 0 ? (_getenv(16) | 0) != 0 : 0) {
  _puts(56) | 0;
  _puts(40) | 0;
  return 0;
 }
 _puts(24) | 0;
 _puts(40) | 0;
 return 0;
['--profiling', '-g2'] 1
1 == 1
function _main(i1, i2) {
 i1 = i1 | 0;
 i2 = i2 | 0;
 if ((_getenv(8) | 0) != 0 ? (_getenv(16) | 0) != 0 : 0) {
  _puts(56) | 0;
  _puts(40) | 0;
  return 0;
 }
 _puts(24) | 0;
 _puts(40) | 0;
 return 0;
````

Personally, I think the newer JS is better than the old JS for the failing test case (from previous comment).

**Comment by kripken** [comment]
Looks like you can reduce that `3` to `2` to make it pass. I agree it's nicer code now. Appears that we didn't split the block before due to the extra stack code at the end, but now we do.

File: lib/Target/JSBackend/AllocaManager.h [hunk 1]
```diff
@@ -140,7 +140,7 @@ class AllocaManager {
   void computeInterBlockLiveness();
   void computeIntraBlockLiveness();
   void computeRepresentatives();
-  void computeFrameOffsets();
+  bool computeFrameOffsets();
```
kripken: comment on what is returned, since it's not obvious

lib/Target/JSBackend/AllocaManager.h
```diff
@@ -148,7 +148,7 @@ class AllocaManager {
   AllocaManager();
 
   /// Analyze the given function and prepare for getRepresentative queries.
```
kripken: comment on what is returned

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1947,6 +1950,9 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
       return;
     }
 
+    // We have a dynamic alloca, so teh stack will be bumped.
```
kripken: teh => the

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1947,6 +1950,9 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
...
+    StackBumped = true;
```
kripken: why can't we move this check higher up, so we catch both static and dynamic allocs here, and avoid needing to change AllocaManager at all?

waywardmonkeys:
Both pull requests (here and an in ``emscripten``) are updated now.

@kripken asked why I needed to structure things the way I did with the changes to ``AllocaManager``. That was the result of an earlier way that I was handling things by having code inside ``getStackBump`` which necessitated that.  However, when I fixed other things in my approach, I missed that I could at that point simplify things a lot.

*Note: I left a comment in there about a remaining bug (that isn't related to my changes). When an alloca is converted to nativized variables, the code is still allocating static stack space for it in the frame. This could be improved as it isn't necessary and that would, in turn, let us get rid of some more stack restorations.*

kripken:
Surely you also want to not emit the `sp = STACKTOP; STACKTOP += N` code as well? I don't see that here.,

waywardmonkeys:
@kripken that is only emitted when it is needed ... when there's a frame offset or a dynamic alloca ... there's nothing to not emit in the case of there not being stack allocations.

kripken:
Ok, cool.

kripken:
This broke `other.test_emterpreter`.

waywardmonkeys:
Do you have any details?  That test is skipped when using node which is why it didn't fail for me.

kripken:
It emits

````
log test /media/alon/2f9a30d7-6124-42d9-87c5-3c80cb70ec54/home/alon/Dev/emscripten/tests/fannkuch.cpp 234
  seen 230
ERROR

======================================================================
ERROR: test_emterpreter (test_other.other)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/media/alon/2f9a30d7-6124-42d9-87c5-3c80cb70ec54/home/alon/Dev/emscripten/tests/test_other.py", line 4567, in test_emterpreter
    do_log_test(path_from_root('tests', 'fannkuch.cpp'), 234, 'fannkuch_worker')
  File "/media/alon/2f9a30d7-6124-42d9-87c5-3c80cb70ec54/home/alon/Dev/emscripten/tests/test_other.py", line 4564, in do_log_test
    assert expected == seen or seen in expected, ['expect', expected, 'but see', seen]
TypeError: argument of type 'int' is not iterable

----------------------------------------------------------------------
Ran 1 test in 42.179s

FAILED (errors=1)
````

kripken:
The failure is a bug with the error reporting in case of a failure. What's going on is that it expects 234 instructions, but sees 230. I.e. your patch shortened the function, I guess due to removing the stack overhead instructions. I'll do some more checking then update the test.

kripken:
Done.

kripken:
This also seems to have broken `other.test_simplify_ifs`. Which is strange since I thought the other pull addressed that...

waywardmonkeys:
That passes for me here... as does a full run of the entire core test suite (using node).

kripken:
I updated it, it's similar to your change but in another place. Looks similarly justified. But now I wonder if it will fail on your machine. Perhaps make sure you cleared the cache before testing again?

PR #65, opened 2015-03-09
Make memcpy / memset inlining configurable

waywardmonkeys:
These settings control how memcpy and memset are either unrolled,
turned into a loop, or emitted as an out-of-line function call.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -93,6 +93,21 @@ GlobalBase("emscripten-global-base",
            cl::desc("Where global variables start out in memory (see emscripten GLOBAL_BASE option)"),
            cl::init(8));
 
+static cl::opt<int>
+MemWriteLoopMax("emscripten-mem-write-loop-max",
+                cl::desc("The count of memcpy or memset steps that should be done as a loop."),
+                cl::init(128));
+
+static cl::opt<int>
+MemUnrollLoopMax("emscripten-mem-unroll-loop-max",
+                 cl::desc("The count of memcpy or memset steps that should be unrolled."),
+                 cl::init(8));
+
+static cl::opt<int>
+PreferSmallerJS("emscripten-prefer-smaller-js",
+                cl::desc("The priority of generating smaller JS. 0 is off, 1 is somewhat, 2 is max."),
+                cl::init(0));
+
 
 extern "C" void LLVMInitializeJSBackendTarget() {
   // Register the target.
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -308,9 +308,6 @@ CMPXCHG_HANDLER(llvm_nacl_atomic_cmpxchg_i8);
 CMPXCHG_HANDLER(llvm_nacl_atomic_cmpxchg_i16);
 CMPXCHG_HANDLER(llvm_nacl_atomic_cmpxchg_i32);
 
-#define UNROLL_LOOP_MAX 8
-#define WRITE_LOOP_MAX 128
-
 DEF_CALL_HANDLER(llvm_memcpy_p0i8_p0i8_i32, {
   if (CI) {
     ConstantInt *AlignInt = dyn_cast<ConstantInt>(CI->getOperand(3));
@@ -319,7 +316,7 @@ DEF_CALL_HANDLER(llvm_memcpy_p0i8_p0i8_i32, {
       if (LenInt) {
         // we can emit inline code for this
         unsigned Len = LenInt->getZExtValue();
-        if (Len <= WRITE_LOOP_MAX) {
+        if (Len <= MemWriteLoopMax) {
           unsigned Align = AlignInt->getZExtValue();
           if (Align > 4) Align = 4;
           else if (Align == 0) Align = 1; // align 0 means 1 in memcpy and memset (unlike other places where it means 'default/4')
@@ -334,7 +331,7 @@ DEF_CALL_HANDLER(llvm_memcpy_p0i8_p0i8_i32, {
             // handle as much as we can in the current alignment
             unsigned CurrLen = Align*(Len/Align);
             unsigned Factor = CurrLen/Align;
-            if (Factor <= UNROLL_LOOP_MAX) {
+            if (Factor <= MemUnrollLoopMax) {
               // unroll
               for (unsigned Offset = 0; Offset < CurrLen; Offset += Align) {
                 std::string Add = "+" + utostr(Pos + Offset);
@@ -368,7 +365,7 @@ DEF_CALL_HANDLER(llvm_memset_p0i8_i32, {
         if (ValInt) {
           // we can emit inline code for this
           unsigned Len = LenInt->getZExtValue();
-          if (Len <= WRITE_LOOP_MAX) {
+          if (Len <= MemWriteLoopMax) {
             unsigned Align = AlignInt->getZExtValue();
             unsigned Val = ValInt->getZExtValue();
             if (Align > 4) Align = 4;
@@ -388,7 +385,7 @@ DEF_CALL_HANDLER(llvm_memset_p0i8_i32, {
                 FullVal |= Val;
               }
               unsigned Factor = CurrLen/Align;
-              if (Factor <= UNROLL_LOOP_MAX) {
+              if (Factor <= MemUnrollLoopMax) {
                 // unroll
                 for (unsigned Offset = 0; Offset < CurrLen; Offset += Align) {
                   std::string Add = "+" + utostr(Pos + Offset);


waywardmonkeys:
I will be following this up shortly with a patch on the emscripten side ...

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -93,6 +93,21 @@ GlobalBase("emscripten-global-base",
...
+
+static cl::opt<int>
+MemUnrollLoopMax("emscripten-mem-unroll-loop-max",
+                 cl::desc("The count of memcpy or memset steps that should be unrolled."),
+                 cl::init(8));
+
+static cl::opt<int>
+PreferSmallerJS("emscripten-prefer-smaller-js",
+                cl::desc("The priority of generating smaller JS. 0 is off, 1 is somewhat, 2 is max."),
+                cl::init(0));
```
sunfishcode: Would it make sense to make more use of clang's -Os and -Oz options, or is there a need for Emscripten to invent its own?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -93,6 +93,21 @@ GlobalBase("emscripten-global-base",
...
 
+static cl::opt<int>
+MemWriteLoopMax("emscripten-mem-write-loop-max",
+                cl::desc("The count of memcpy or memset steps that should be done as a loop."),
+                cl::init(128));
+
+static cl::opt<int>
+MemUnrollLoopMax("emscripten-mem-unroll-loop-max",
+                 cl::desc("The count of memcpy or memset steps that should be unrolled."),
+                 cl::init(8));
```
sunfishcode: No other backends in LLVM have options like this. Would it be sufficient for us to use predefined values for these if we make them sensitive to the optsize/minsize (-Os/-Oz) flags?

lib/Target/JSBackend/JSBackend.cpp
waywardmonkeys: That's a _fascinating_ question!  The answer took me a while to work out before I went down this path.

``-Os`` and ``-Oz`` are not handled the same as the other ``-O#`` parameters ... what happens is that when ``-Os`` and ``-Oz`` are set, then function attributes get set to specify that that function should be compiled for size or min-size.

These function attributes is how all of the other targets handle ``-Os`` and ``-Oz``. In the emscripten case, we often have people (like us) who are building with clang to generate .bc files and then pumping that .bc content through emcc to generate the JS, so the normal model of handling this doesn't work out so well.

You can see the clang side of this in ``tools/clang/lib/CodeGen/CGCall.cpp``:

````c++
  if (CodeGenOpts.OptimizeSize)
    FuncAttrs.addAttribute(llvm::Attribute::OptimizeForSize);
  if (CodeGenOpts.OptimizeSize == 2)
    FuncAttrs.addAttribute(llvm::Attribute::MinSize);
````

No other target or LLVM-level (not Clang-level) code that I'm aware of was looking at the command line arguments since those were handled in Clang.

I also thought that it was useful to let these be tuned rather than hard-coded ...

lib/Target/JSBackend/JSBackend.cpp
waywardmonkeys: I'm not sure that any other backend does something quite like the code expansion that happens to ``memcpy`` and ``memset`` ... it is similar to inlining though, which does have some switches to help control it.

waywardmonkeys: This inlining of ``memcpy`` and ``memset`` does result in absolutely terrible codegen in terms of code size. It is a visible source of bloat for us and one that we regularly work around at the C++ level.

lib/Target/JSBackend/JSBackend.cpp
sunfishcode: PreferSmallerJS is not used in this patch. What kinds of things to you anticipate using it for?

Also, yes, the -Os/-Oz options are communicated to the backends via function attributes. This gives programmers the ability to compile some parts of their program in optimize-for-size mode, and other parts in optimize-for-speed mode. This flexibility would be useful for Emscripten as well.

waywardmonkeys: If we could control things so that only auto-generated ``memcpy`` and ``memset`` calls (inserted usually for constructors, copy constructors) got the inlining treatment and any manual call to them was left alone ... that would be great.

Our main issue with this stuff is a result of using ``std::string`` ...

lib/Target/JSBackend/JSBackend.cpp
sunfishcode: LLVM's other backends do do essentially the same kind of inline expansion of memset and memcpy that Emscripten is doing. Take a look at getMaxStoresPerMemcpy and friends in TargetLowering.h, and the code that uses it. Emscripten doesn't use this infrastructure because it doesn't use the SelectionDAG infrastructure, but it'd be nice to behave in a similar manner unless we have a reason to differ.

I totally believe that inlining of memcpy and memset is a significant source of bloat for you, and I'm happy you're helping us fix it. If there's a setting which is better overall, we would like to just change Emscripten to do that by default, since this is likely to affect other Emscripten users too. Exposing Emscripten-specific low-level knobs means that in effect, we're only fixing the problem for users with the time and patience to fiddle with them, so we'd like to avoid them unless there really are use-case-specific concerns in play.

waywardmonkeys: I understand that and agree with the sentiment ... this just seemed like the most straightforward approach. (I maintain another compiler that is pretty large and very old, so I understand the various considerations for maintainability, accessibility, etc.)

I haven't been able to find a way to distinguish between an automatically-inserted call to ``memcpy`` and ``memset`` and one that is manually provided in the source code. Maybe that's a useful change to make upstream with a function attribute, but then it won't land in emscripten for ages (3.7).

lib/Target/JSBackend/JSBackend.cpp
kripken: It would indeed be nice to use `-Os, -Oz` automatically. For example, we recently made those optimizations enable registerizeHarder at link time, as it reduces code size.

What does LLVM do with `-Os,-Oz` in backends? If emcc links to JS with `-Oz`, and we can pass that to the backend, and then tweak memcpy inlining parameters based on that, it seems like the most automatic way to do this.

kripken: Or should we just notice the function attribute? That means we care about `-Oz` during  source compilation, and not link, for purposes of inlining memcpy. Maybe that makes sense though.

sunfishcode: @waywardmonkeys It's unlikely that you'd be able to convince upstream clang/llvm developers to annotate auto-generated memcpy or memset calls as such. They will surely ask: what is it about auto-generated memcpy/memset which makes them different from hand-written calls? Is it having a modestly small constant size and known alignment? If so, then the backend can just look for that and optimize accordingly.

@kripken Yes, clang translates -Os/-Oz into function attributes, so Emscripten should just look for the function attributes.

lib/Target/JSBackend/JSBackend.cpp
sunfishcode: It's unlikely that the clang developers would be amenable to such an attribute; in the clang/llvm world view, it's better to make auto-inserted memcpy/memset look as much like user-inserted memcpy/memset as possible so that they can participate in optimizations in the same way. This may frustrate the user wishing to micro-manage the backend, but it also makes the overall system more predictable and consistent, as two constructs which do essentially the same thing are more likely to be optimized down to the same output code.

It's possible that whatever you want to do for auto-generated memset/memcpy here would be beneficial for any memset/memcpy with a reasonably small constant size and a known useful alignment. So, some questions:
 - Do we need heuristics which are smarter about handling low alignment cases? Right now, the code which decides whether to emit inline code or not just compares the total byte size of the copy with WRITE_LOOP_MAX, rather than taking into consideration how many actual loads and stores would be emitted in the unrolled case.
 - What are the best values for UNROLL_LOOP_MAX and WRITE_LOOP_MAX (or whatever we change it to, in the first question)? And should we replace them with something which can check whether the containing function has optsize/minsize attributes?

kripken:
Ok, we definitely want this functionality, and it sounds like the function attribute thing is the way to go here. So we can have heuristic numbers for how to inline memcpy and memset based on whether the function is -Os or -Oz compiled.

And later I'll do the same for the relooper, on -Os and -Oz methods, we do not want to split nodes. I believe that should always be good for code size, but I'll verify.

kripken:
I made the relooper not split nodes on -Os/-Oz functions. This reduces code size by 0.5% on one of the exceptions testcases.

kripken:
@waywardmonkeys , what do you think about moving forward with this pull, as mentioned earlier, using the function attributes (as we now do for the relooper node splitting)?

kripken:
I still think we should do this: In functions marked optsize or minsize, we should decrease their size as much as we can, in particular by not expanding out memcpy/memset. Anyone opposed?

waywardmonkeys:
I ... think I might be.

kripken:
Enough to explain yourself? :)

waywardmonkeys:
Yes, in a couple of hours. Saturday morning here and I am out for a while.

waywardmonkeys:
I'm concerned that this has a larger impact than would be desired due to the compiler generating calls to ``memset`` & ``memcpy`` for constructors and copy constructors.

I'm working to get actual measurements of this from our codebase.

kripken:
Measurements would be good.

It would only emit such calls if the constructor were inlined into code built with -Os or -Oz, so it seems like we would in fact want the smaller code size there?

waywardmonkeys:
Well, the issue is that it is new semantics for -Os / -Oz in terms of how memcpy / memset are handled.  That is complicated by the fact that we don't (can't) control when memcpy / memset might be called since the compiler changes some other code into a memcpy / memset at times.

I'd rather not make any changes here in the near future, see how things are with the new incoming 3.6 compiler and then look further at what changes are worth making.

waywardmonkeys:
Abandoning.

PR #66, opened 2015-03-11
Nativized variables shouldn't consume stack space.

waywardmonkeys:
Prior to this change, whether or not a variable had been "nativized"
wasn't taken into account when calculating the static space required
for the frame.

Now, they aren't included in that and a minor safety measure is taken
to ensure that the frame offset of a nativized variable isn't requested.

Additionally, the logic for skipping a stack restoration no longer
emits a stack restore when only nativized variables are present within
the frame.

### DIFF
--- a/lib/Target/JSBackend/AllocaManager.cpp
+++ b/lib/Target/JSBackend/AllocaManager.cpp
@@ -371,7 +371,7 @@ void AllocaManager::computeRepresentatives() {
   }
 }
 
-void AllocaManager::computeFrameOffsets() {
+void AllocaManager::computeFrameOffsets(const std::set<const Value*> &nativizedVars) {
   NamedRegionTimer Timer("Compute Frame Offsets", "AllocaManager",
                          TimePassesIsEnabled);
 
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
I'm still running tests on this, but it should be fine for review. It seems to work so far.

kripken:
I believe that nativized variables are only useful in `-O0`. Otherwise LLVM will do it for us. Do you see this as making a difference just in -O0, or also optimized modes?

waywardmonkeys:
Not sure... I was doing it as a cleanup from the previous changes because I was seeing unexpected behavior in the generated JS and figured that it made sense to remove them. I haven't tried building our codebase against this version of fastcomp yet.

waywardmonkeys:
This passed all tests.

lib/Target/JSBackend/AllocaManager.cpp
```diff
@@ -407,19 +407,25 @@ void AllocaManager::computeFrameOffsets() {
   for (SmallVectorImpl<AllocaInfo>::const_iterator I = SortedAllocas.begin(),
        E = SortedAllocas.end(); I != E; ++I) {
     const AllocaInfo &Info = *I;
-    uint64_t NewOffset = RoundUpToAlignment(CurrentOffset, Info.getAlignment());
+    const AllocaInst *AI = Info.getInst();
+
+    // Skip any allocas that have been nativized
+    if (nativizedVars.count(AI)) {
+      StaticAllocas[AI] = StaticAllocation(AI, false);
```
sunfishcode: Would it be possible to avoid creating sentinal StaticAllocation instances for nativized allocas? If there's code that's looking up allocas in StaticAllocas, it could possibly filter out nativized allocas before doing so.

waywardmonkeys: Indeed ... it was a call to ``getRepresentative`` that didn't need to happen.  Fixed and removed the sentinel value.

At the same time, I converted things to use ``isNativizedVar``.

kripken:
I would still like to understand what this wins us, as in my previous comment.

waywardmonkeys:
@kripken How / when / where does LLVM do the nativization at other optimization levels?

If nothing else, this is an easy clean up (and painless) at ``-O0`` and removes a source of confusion that I had when reading generated code.

waywardmonkeys:
Also, some of the "noise" in the commit is adding the ``isNativizedVar`` method as a clean up ...

kripken:
mem2reg I believe, is the pass that replaces allocas with registers (when possible), which is basically that nativization does.

waywardmonkeys:
Abandoning.

PR #67, opened 2015-03-17
Pthreads

juj:
Adds LLVM side support for compiling for the pthreads/atomics support.

This set of commits implements initial pthreads library support for Emscripten code, and allows native code to utilize multithreading with the pthreads API. This backs on experimental research on the JavaScript SharedArrayBuffer and Atomics APIs from https://docs.google.com/document/d/1NDGA_gZJ7M7w1Bh8S0AoDyEqwDdRh4uSoTPSNn77PFk/edit?usp=sharing , and as such is preliminary and unsupported in other browsers except current Firefox Nightly builds. By default pthreads support is disabled, and the feature must be explicitly enabled by specifying the compiler flag -s USE_PTHREADS=1 and the linker flag -lpthread.

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -588,13 +588,13 @@ DEF_CALL_HANDLER(emscripten_atomic_load_u32, {
 DEF_CALL_HANDLER(emscripten_atomic_load_f32, {
   // TODO: If https://bugzilla.mozilla.org/show_bug.cgi?id=1131613 is implemented, we could use the commented out version. Until then,
   // we must emulate manually.
-  return getAssign(CI) + "__Atomics_load_f32_emulated(" + getAddressAsString(CI->getOperand(0), 2) + ")";
+  return getAssign(CI) + (PreciseF32 ? "Math_fround(" : "+") + "__Atomics_load_f32_emulated(" + getAddressAsString(CI->getOperand(0), 2) + (PreciseF32 ? "))" : ")");
 //  return getAssign(CI) + "Atomics_load(HEAPF32, " + getAddressAsString(CI->getOperand(0), 2) + ")";
 })
 DEF_CALL_HANDLER(emscripten_atomic_load_f64, {
   // TODO: If https://bugzilla.mozilla.org/show_bug.cgi?id=1131624 is implemented, we could use the commented out version. Until then,
   // we must emulate manually.
-  return getAssign(CI) + "__Atomics_load_f64_emulated(" + getAddressAsString(CI->getOperand(0), 3) + ")";
+  return getAssign(CI) + "+__Atomics_load_f64_emulated(" + getAddressAsString(CI->getOperand(0), 3) + ")";
 //  return getAssign(CI) + "Atomics_load(HEAPF64, " + getAddressAsString(CI->getOperand(0), 3) + ")";
 })
 
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -318,7 +318,7 @@ DEF_CALL_HANDLER(llvm_nacl_atomic_store_i32, {
 DEF_CALL_HANDLER(name, { \
   const Value *P = CI->getOperand(0); \
   if (EnablePthreads) { \
-    return getAssign(CI) + "Atomics_compareExchange(" HeapName ", " + getAddressAsString(CI->getOperand(0), 2) + ", " + getValueAsStr(CI->getOperand(1)) + ", " + getValueAsStr(CI->getOperand(2)) + ")"; \
+    return getAssign(CI) + "Atomics_compareExchange(" HeapName ", " + getShiftedPtr(CI->getOperand(0), 4) + ", " + getValueAsStr(CI->getOperand(1)) + ", " + getValueAsStr(CI->getOperand(2)) + ")"; \
   } else { \
     return getLoad(CI, P, CI->getType(), 0) + ';' + \
              "if ((" + getCast(getJSName(CI), CI->getType()) + ") == " + getValueAsCastParenStr(CI->getOperand(1)) + ") " + \
@@ -554,31 +554,22 @@ DEF_CALL_HANDLER(emscripten_asm_const_double, {
   return getAssign(CI) + getCast(handleAsmConst(CI), Type::getDoubleTy(CI->getContext()));
 })
 
-std::string getAddressAsString(const Value *Ptr, int shift) {
-  Type *t = cast<PointerType>(Ptr->getType())->getElementType();
-  if (const GlobalVariable *GV = dyn_cast<GlobalVariable>(Ptr)) {
-    return relocateGlobal(utostr(getGlobalAddress(GV->getName().str()) >> shift));
-  } else {
-    return getValueAsStr(Ptr) + ">>" + utostr(shift);
-  }
-}
-
 /* TODO: Uncomment once https://bugzilla.mozilla.org/show_bug.cgi?id=1141986 is implemented!
 
 DEF_CALL_HANDLER(emscripten_atomic_exchange_u8, {
   return getAssign(CI) + "Atomics_exchange(HEAP8, " + getValueAsStr(CI->getOperand(0)) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
 })
 DEF_CALL_HANDLER(emscripten_atomic_exchange_u16, {
-  return getAssign(CI) + "Atomics_exchange(HEAP16, " + getAddressAsString(CI->getOperand(0), 1) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
+  return getAssign(CI) + "Atomics_exchange(HEAP16, " + getShiftedPtr(CI->getOperand(0), 2) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
 })
 DEF_CALL_HANDLER(emscripten_atomic_exchange_u32, {
-  return getAssign(CI) + "Atomics_exchange(HEAP32, " + getAddressAsString(CI->getOperand(0), 2) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
+  return getAssign(CI) + "Atomics_exchange(HEAP32, " + getShiftedPtr(CI->getOperand(0), 4) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
 })
 DEF_CALL_HANDLER(emscripten_atomic_exchange_f32, {
-  return getAssign(CI) + "Atomics_exchange(HEAPF32, " + getAddressAsString(CI->getOperand(0), 2) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
+  return getAssign(CI) + "Atomics_exchange(HEAPF32, " + getShiftedPtr(CI->getOperand(0), 4) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
 })
 DEF_CALL_HANDLER(emscripten_atomic_exchange_f64, {
-  return getAssign(CI) + "Atomics_exchange(HEAPF64, " + getAddressAsString(CI->getOperand(0), 3) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
+  return getAssign(CI) + "Atomics_exchange(HEAPF64, " + getShiftedPtr(CI->getOperand(0), 8) + ", " + getValueAsStr(CI->getOperand(1)) + ")";
 })
 */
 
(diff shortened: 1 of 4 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -855,15 +927,36 @@ std::string JSWriter::getLoad(const Instruction *I, const Value *P, Type *T, uns
   return text;
 }
 
+static const std::string AtomicsStore = "Atomics_store(";
```
kripken: Is this a useful optimization here? Is there a specific reason why it is done on this string and not others?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -855,15 +927,36 @@ std::string JSWriter::getLoad(const Instruction *I, const Value *P, Type *T, uns
...
+      text = getPtrUse(P) + " = " + VS;
+    }
     if (Alignment == 536870912) text += "; abort() /* segfault */";
   } else {
     // unaligned in some manner
+
+    if (EnablePthreads && cast<StoreInst>(I)->isVolatile()) {
+      errs() << "emcc: warning: unable to implement unaligned volatile store as atomic in " << I->getParent()->getParent()->getName() << ":" << *I << " | ";
+      emitDebugInfo(errs(), I);
+      errs() << "\n";
```
kripken: shouldn't we abort here?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2148,31 +2241,88 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
                                     getValueAsStr(I->getOperand(2));
     break;
   }
+  case Instruction::ExtractValue: {
```
kripken: why is this here? I thought the pnacl passes lower out ExtractValue?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2148,31 +2241,88 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
...
+    break;
+  }
+  case Instruction::AtomicCmpXchg: {
+    const AtomicCmpXchgInst *cxi = cast<AtomicCmpXchgInst>(I);
+    const Value *P = I->getOperand(0);
+    const char *HeapName;
+    std::string Index = getHeapNameAndIndex(P, &HeapName);
+    std::string Assign = getJSName(I);
+    if (EnablePthreads) {
+      Code << Assign << "_0 = Atomics_compareExchange(" << HeapName << ", " << Index << ", " << getValueAsStr(I->getOperand(1)) << ", " << getValueAsStr(I->getOperand(2)) << ");\n";
```
kripken: i think we might need to declare these variables in `UsedVars` so they are in the `var` statement?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2148,31 +2241,88 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
...
+        Code << Assign << "__" << atomicFunc << HeapName << "_emulated(" << getByteAddressAsStr(P) << ", " << VS << ")"; break;
+      } else {
+        Code << Assign << atomicFunc << "(" << HeapName << ", " << Index << ", " << VS << ")"; break;
+      }
+    } else {
+      Code << getLoad(rmwi, P, I->getType(), 0) << ';';
+      // Most bitcasts are no-ops for us. However, the exception is int to float and float to int
+      switch (rmwi->getOperation()) {
+        case AtomicRMWInst::Xchg: Code << getStore(rmwi, P, I->getType(), VS, 0); break;
+        case AtomicRMWInst::Add: Code << getStore(rmwi, P, I->getType(), "((" + getJSName(I) + '+' + VS + ")|0)", 0); break;
```
kripken: why did this alignment break? (`Code << `)

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2148,31 +2241,88 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
...
+        case AtomicRMWInst::Max:
+        case AtomicRMWInst::Min:
+        case AtomicRMWInst::UMax:
+        case AtomicRMWInst::UMin:
+        case AtomicRMWInst::BAD_BINOP: llvm_unreachable("Bad atomic operation");
+      }
     }
     break;
   }
   case Instruction::Fence: // no threads, so nothing to do here
```
kripken: can move this comment down

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2490,6 +2640,7 @@ void JSWriter::printModuleBody() {
   assert(GlobalData32.size() == 0 && GlobalData8.size() == 0); // FIXME when we use optimal constant alignments
 
   // TODO fix commas
+  Out << "if (!ENVIRONMENT_IS_PTHREAD) {\n";
```
kripken: various things in emscripten look for the `memory initializer` command, so they can modify it, like turning the memory initalizer into a binary mem init file, and the emterpreter which adds more binary data into the memory initalizer (the bytecode). would those not work when pthreads are enabled?

I assume `ENVIRONMENT_IS_PTHREAD` is in the worker threads only?

lib/Transforms/NaCl/RewriteAtomics.cpp
```diff
@@ -357,6 +357,7 @@ void AtomicVisitor::visitAtomicRMWInst(AtomicRMWInst &I) {
 ///   %success = icmp eq %old, %val
 /// Note: weak is currently dropped if present, the cmpxchg is always strong.
 void AtomicVisitor::visitAtomicCmpXchgInst(AtomicCmpXchgInst &I) {
+  return; // XXX EMSCRIPTEN
```
kripken: why?

lib/Target/JSBackend/JSBackend.cpp
juj: Hmm, let me just remove this, it is a leftover from previous form of code. I hope compilers are smart to optimize this on their own anyways.

lib/Target/JSBackend/JSBackend.cpp
juj: This is better to be a warning, since this situation can occur in perfectly valid C/C++ code, it is just that we are unable to give strong guarantees about `volatile` being visible to other threads in this case.

lib/Target/JSBackend/JSBackend.cpp
juj: I had to implement this to add support for the LLVM compare-and-exchange operation. That operation generates a structure pair of (oldValue, boolean_succeeded), and I started getting errors about an invalid instruction if I did not implement this myself.

lib/Target/JSBackend/JSBackend.cpp
juj: Oh! Thanks, added.

lib/Target/JSBackend/JSBackend.cpp
juj: Restored whitespace alignment.

lib/Target/JSBackend/JSBackend.cpp
juj: Done.

lib/Target/JSBackend/JSBackend.cpp
juj: Hmm, I don't know. I think I have run all the tests and everything pass, so should be good. Actually, I see that I should not be emitting this `if (!ENVIRONMENT_IS_PTHREAD)` statement at all if `EnablePthreads` is false, so updated this to only conditionally emit this check. I want the JS variable `ENVIRONMENT_IS_PTHREAD` to not exist at all in the generated code if it was not built with pthreads enabled, so that existing compilation will not be affected/disrupted by added code size by this feature.

lib/Transforms/NaCl/RewriteAtomics.cpp
juj: This follows the same pattern as the other functions in this file which had already been disabled for Emscripten purposes: `AtomicVisitor::visitAtomicRMWInst`, `AtomicVisitor::visitStoreInst`, `AtomicVisitor::visitLoadInst`, `AtomicVisitor::visitFenceInst`. I think the purpose of these functions is to change LLVM operations to NaCl-specific intrinsic function call(?) operations, which we don't use, so these are disabled.

juj:
Addressed the review comments at https://github.com/juj/emscripten-fastcomp/compare/a5b54c1...0bc3f1a .

lib/Target/JSBackend/JSBackend.cpp
kripken: Unaligned stores are almost always due to undefined behavior, aren't they? I wonder if we shouldn't give an error and let people fix that.

lib/Target/JSBackend/JSBackend.cpp
kripken: In that case, I think we need to improve the lowering pass. The approach of adding `_0`, `_1` to variables is not very robust.

It's possible the passes are already fixed btw, in latest pnacl, which is on incoming? If not, we can file a bug on pnacl.

lib/Target/JSBackend/JSBackend.cpp
kripken: You didn't run the whole test suite with pthreads enabled, I assume?

If stuff like the emterpreter fails with pthreads, that's not a reason to not land this, but we should run the full test suite with pthreads enabled, see what breaks, and give an error if those are attempted to be used together with pthreads.

lib/Transforms/NaCl/RewriteAtomics.cpp
kripken: I see. However we do need to get the structs legalized out, so we don't have to implement ExtractValue etc. in the backend. At least that's how we've done things so far, and avoided `_0,_1` variables as a hack for them. Perhaps @sunfishcode has thoughts here?

sunfishcode: I don't have any great ideas either. I actually think the _0,_1 code is fine, provided they don't collide with other names somehow. Although, if we do this, we should add an assert to the extractvalue code asserting that the input is actually a cmpxchg, since the code there isn't prepared to handle any other kind of extractvalue.

kripken: Aside from collisions, another concern is the way we look through bitcasts, do nativization, etc.

Overall I think it would be simpler to do a little legalization pass that lowers cmpxchng like we lower i64, setjmp, etc. - to multiple target variables using 2 or so emscripten intrinsics, then just trivially implementing those.

lib/Target/JSBackend/JSBackend.cpp
juj: Well, compilers do have specific attributes to specify custom alignment, so in practice the code can be legal. The `volatile` keyword in C89 means

	An object that has volatile-qualified type may be modified in ways unknown to
	the implementation or have other unknown side effects. Therefore any
	expression referring to such an object shall be evaluated strictly according
	to the rules of the abstract machine, as described in 2.1.2.3
	(http://port70.net/~nsz/c/c89/c89-draft.html#2.1.2.3). Furthermore, at every
	sequence point the value last stored in the object shall agree with that
	prescribed by the abstract machine, except as modified by the unknown factors
	mentioned previously. A volatile declaration may be used to describe an object
	corresponding to a memory-mapped input/output port or an object accessed by an
	asynchronously interrupting function. Actions on objects so declared shall not
	be ``optimized out'' by an implementation or reordered except as permitted by
	the rules for evaluating expressions. What constitutes an access to an object
	that has volatile-qualified type is implementation-defined.

So a volatile load or store is not defined to be a multithreaded atomic operation, nor is it required to be an aligned operation, it is just intended to be one that is visible immediately without optimizations. In the JS SharedArrayBuffer spec, we don't have an API to tell the JS engine to make an operation immediately visible (== force omitting optimizations), so we do something stronger, i.e. make the operations atomic, which will also guarantee the operations to not be optimized out. However if the operation was not aligned, then we don't have this option, so we just generate a standard unaligned load or store. If there exists an atomic operation immediately later in the code (as there should), everything will most likely be fine and the code will function correctly even though the operation was not aligned, so therefore I'd like to retain this just as a warning.

kripken: ok.

lib/Target/JSBackend/CallHandlers.h
```diff
@@ -549,6 +549,162 @@ DEF_CALL_HANDLER(emscripten_asm_const_double, {
   return getAssign(CI) + getCast(handleAsmConst(CI), Type::getDoubleTy(CI->getContext()));
 })
 
+std::string getAddressAsString(const Value *Ptr, int shift) {
+  Type *t = cast<PointerType>(Ptr->getType())->getElementType();
+  if (const GlobalVariable *GV = dyn_cast<GlobalVariable>(Ptr)) {
+    return utostr(getGlobalAddress(GV->getName().str()) >> shift);
```
kripken: `getGlobalAddress` returns a number. that number does not include a relocation offset, which we need in shared modules. `relocateGlobal` will do a relocation for you (it operates on a string, so need to take into account the shift on the extra part here).

lib/Target/JSBackend/CallHandlers.h
```diff
@@ -549,6 +549,162 @@ DEF_CALL_HANDLER(emscripten_asm_const_double, {
...
+  return getAssign(CI) + "Atomics_compareExchange(HEAP8, " + getValueAsStr(CI->getOperand(0)) + ", " + getValueAsStr(CI->getOperand(1)) + ", " + getValueAsStr(CI->getOperand(2)) + ")";
+})
+DEF_CALL_HANDLER(emscripten_atomic_cas_u16, {
+  return getAssign(CI) + "Atomics_compareExchange(HEAP16, " + getAddressAsString(CI->getOperand(0), 1) + ", " + getValueAsStr(CI->getOperand(1)) + ", " + getValueAsStr(CI->getOperand(2)) + ")";
+})
+DEF_CALL_HANDLER(emscripten_atomic_cas_u32, {
+  return getAssign(CI) + "Atomics_compareExchange(HEAP32, " + getAddressAsString(CI->getOperand(0), 2) + ", " + getValueAsStr(CI->getOperand(1)) + ", " + getValueAsStr(CI->getOperand(2)) + ")";
+})
+DEF_CALL_HANDLER(emscripten_atomic_cas_f32, {
+  errs() << "emcc: warning: float32 compare-and-swap is not supported!" << CI->getParent()->getParent()->getName() << ":" << *CI << "\n";
```
kripken: we have a `prettyWarning` method

kripken: this and similar warnings will get printed a lot, i worry? or in practice are there just a few? seems like it could flood the output. one option is to put it behind the `EmscriptenAssertions` option.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -840,18 +847,91 @@ std::string JSWriter::getIMul(const Value *V1, const Value *V2) {
   return "Math_imul(" + getValueAsStr(V1) + ", " + getValueAsStr(V2) + ")|0"; // unknown or too large, emit imul
 }
 
+/// Given a pointer to memory, returns the HEAP object and index to that object that is used to access that memory.
```
kripken: comment should probably be in the definition, not the implementation?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -840,18 +847,91 @@ std::string JSWriter::getIMul(const Value *V1, const Value *V2) {
...
+/// @param Ptr [in] The heap object.
+/// @param HeapName [out] Receives the name of the HEAP object used to perform the memory acess.
+/// @return The index to the heap HeapName for the memory access.
+std::string JSWriter::getHeapNameAndIndex(const Value *Ptr, const char **HeapName)
+{
+  Type *t = cast<PointerType>(Ptr->getType())->getElementType();
+  unsigned Bytes = DL->getTypeAllocSize(t);
+
+  if (const GlobalVariable *GV = dyn_cast<GlobalVariable>(Ptr)) {
+    unsigned Addr = getGlobalAddress(GV->getName().str());
```
kripken: this hits the same issue with `relocateGlobal` being needed. the deeper issue is that this duplicates code with `getHeapAccess` and `getPtrUse`. If those are not sufficient, we could add `getHeap` which just returns `HEAPF64` etc., but leaves the shifts etc. to another method, and reuse that method from `getPtrUse` and `getHeapAccess`.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -840,18 +847,91 @@ std::string JSWriter::getIMul(const Value *V1, const Value *V2) {
...
+        if (t->isIntegerTy() || t->isPointerTy()) *HeapName = "HEAP32";
+        else *HeapName = "HEAPF32";
+        return Index + ">>2";
+      case 2: *HeapName = "HEAP16"; return Index + ">>1";
+      case 1: *HeapName = "HEAP8"; return Index;
+    }
+  }
+}
+
+std::string JSWriter::getByteAddressAsStr(const Value *Ptr)
```
kripken: isn't this the same as calling `getValueAsStr`? that will end up doing the same via a call to `getConstant` (which would also get the relocation).

lib/Target/JSBackend/JSBackend.cpp
kripken: what is the status of this? we have a new pnacl merge on incoming, maybe they can legalize this now?

the `_0` etc. approach is something we should avoid. if pnacl can't do it, we may need to add a lowering pass

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2265,31 +2368,96 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
...
+    Code << Assign << getValueAsStr(I->getOperand(0)) << "_" << evi->idx_begin()[0];
+    break;
+  }
+  case Instruction::AtomicCmpXchg: {
+    const AtomicCmpXchgInst *cxi = cast<AtomicCmpXchgInst>(I);
+    const Value *P = I->getOperand(0);
+    const char *HeapName;
+    std::string Index = getHeapNameAndIndex(P, &HeapName);
+    std::string Assign = getJSName(I);
+    UsedVars[Assign + "_0"] = I->getOperand(1)->getType();
```
kripken: same issue

lib/Transforms/NaCl/RewriteAtomics.cpp
kripken: as mentioned above, i still think this way.

lib/Target/JSBackend/CallHandlers.h
juj: Ok, fixed. Although note that threads and shared libraries is not supported at present, that is left for later.

lib/Target/JSBackend/CallHandlers.h
juj: Actually, float cas is not used anywhere nor supported, so I'll just remove these.

lib/Target/JSBackend/JSBackend.cpp
juj: Moved the comment to the function declaration.

lib/Target/JSBackend/JSBackend.cpp
juj: Added a commit to remove the code duplication with these functions.

lib/Target/JSBackend/JSBackend.cpp
juj: Looks like so. Removed getByteAddressAsStr().

lib/Target/JSBackend/JSBackend.cpp
juj: Indeed looks like PNaCl works here now, so changed to use that instead.

lib/Target/JSBackend/JSBackend.cpp
juj: Removed this code.

lib/Transforms/NaCl/RewriteAtomics.cpp
juj: Removed this code in favor of the PNaCl intrinsic, so the issue should not occur anymore.

juj:
Addressed the review comments and rebased on top of the latest. This should hopefully look simpler now?

lib/Target/JSBackend/CallHandlers.h
```diff
@@ -549,6 +554,154 @@ DEF_CALL_HANDLER(emscripten_asm_const_double, {
   return getAssign(CI) + getCast(handleAsmConst(CI), Type::getDoubleTy(CI->getContext()));
 })
 
+std::string getAddressAsString(const Value *Ptr, int shift) {
```
kripken: Let's call this `getShiftedPtr`, as it works on pointers, and it doesn't return just a string, but specifically it returns a shifted result (so it isn't the real pointer/address anymore).

Also, please move this to right before `getPtrUse` in the main `JSBackend.cpp` file. Then the distinction between the two methods is clear.

And once at that location, it looks like we can remove `getHeapAccess`, and instead implement `getPtrUse` using the new `getShiftedPtr` (the only extra thing needed is `HEAP?[` on the left and `]` on the right, can do that in `getPtrUse`, or alternatively leave `getHeapAccess` but without the shifting inside it, so all it does is add the `HEAP?[` and `]`; I am ok with either).

kripken: Oh, I see there is `getHeapName`, so can use that for the `HEAP?` part, leaving just `[` and `]`.

kripken: Or actually instead of everything I said, this method could be implemented using the `getHeapNameAndIndex` family of methods.

kripken: Anyhow, sorry to ramble here, but in summary

1. I think `getShiftedPtr` is a better name
2. I'd prefer this to be implemented using other methods, without a special check on `GlobalVariable`, and without the need to `relocateGlobal`, as the other methods already do both.

juj: Ok, updated the pull request to rename this to `getShiftedPtr` and to reuse the existing code as much as possible.

kripken:
Looks nice, thanks.

One last thing, just to be sure, please test compilation time before and after this pull. I'm not too worried, but still this changes a lot of the string handling code in emitting pointer uses in the backend. A quick test on a large codebase of js backend time, with and without this pull, should be enough to check.

If that is ok, then lgtm.

juj:
I rebuilt Unity3D three times with trunk and both pthreads PRs applied, with the following results:

    incoming:
    emscript: llvm backend took 11.1803879738 seconds
    emscript: llvm backend took 10.9915230274 seconds
    emscript: llvm backend took 10.9425158501 seconds

    pthreads:
    emscript: llvm backend took 11.1047959328 seconds
    emscript: llvm backend took 11.1267051697 seconds
    emscript: llvm backend took 11.0028831959 seconds

so the differences do look quite negligible. I'll go ahead and merge these in, if everything is set?

kripken:
Sure!

PR #69, opened 2015-03-31
Virtual issues

waywardmonkeys:


lib/Target/JSBackend/JSTargetTransformInfo.cpp
```diff
@@ -55,24 +55,26 @@ class JSTTI : public ImmutablePass, public TargetTransformInfo {
...
+                                  OperandValueProperties Opd1PropInfo = OP_None,
+                                  OperandValueProperties Opd2PropInfo = OP_None) const override;
 
-  virtual unsigned getVectorInstrCost(unsigned Opcode, Type *Val,
-                                      unsigned Index = -1) const;
+  unsigned getVectorInstrCost(unsigned Opcode, Type *Val,
+                              unsigned Index = -1) const override;
 
-  virtual void getUnrollingPreferences(Loop *L, UnrollingPreferences &UP) const;
+  void getUnrollingPreferences(const Function *F, Loop *L, UnrollingPreferences &UP) const override;
```
kripken: this looks like the right signature in the headers, but i wonder how it worked without this change, so i seem to be misunderstanding something

cc @sunfishcode

sunfishcode: It looks like the signature in the base class changed in LLVM 3.6.

kripken: yes, but how did this work without @waywardmonkeys 's fix? was it overloading the method name, but the method wasn't actually being called? if so it sounds like something we should have tests for.

waywardmonkeys: The compiler was emitting warnings for both of the changed signatures. My change to use ``override`` makes them into compilation errors.

sunfishcode: Yes, I assume it wasn't being called. These are heuristics, so they aren't strictly necessary. Also, yes, we should ideally have tests for these. The code that's there right now is just some initial guesswork, something we can start experimenting with.

PR #72, opened 2015-04-03
Remove raw_ostream localmod

jfbastien:
This was added by @juj in e8fdaf5a2f09e335bdc5a6689b298e54337cc737 to fix the VS2010 build. It's either not needed anymore because the diff from LLVM upstream on this file is empty, or it's instead needed in upstream LLVM (I doubt it, since LLVM does build correcting on MSVC).

### DIFF
--- a/lib/IR/Module.cpp
+++ b/lib/IR/Module.cpp
@@ -22,7 +22,6 @@
 #include "llvm/IR/GVMaterializer.h"
 #include "llvm/IR/InstrTypes.h"
 #include "llvm/IR/LLVMContext.h"
-#include "llvm/Support/raw_ostream.h" // XXX Emscripten TODO: Move to PNacl upstream.
 #include "llvm/IR/LeakDetector.h"
 #include "llvm/Support/Dwarf.h"
 #include "llvm/Support/Path.h"


kripken:
@juj , maybe you want to test this one locally before merging?

jfbastien:
Yes, please test locally. AFAICT it should be fine, I'm assuming that we removed the localmod that led to this ~1 year old patch.

juj:
Doing builds now..

juj:
The code does build with this ok, so merging.

However, both with and without, I'm getting a linker error `54>LLVMJSBackendCodeGen.lib(JSBackend.obj) : error LNK2019: unresolved external symbol "class llvm::FunctionPass * __cdecl llvm::createExpandInsertExtractElementPass(void)" (?createExpandInsertExtractElementPass@llvm@@YAPEAVFunctionPass@1@XZ) referenced in function "public: virtual bool __cdecl llvm::JSTargetMachine::addPassesToEmitFile(class llvm::legacy::PassManagerBase &,class llvm::formatted_raw_ostream &,enum llvm::TargetMachine::CodeGenFileType,bool,void const *,void const *)" (?addPassesToEmitFile@JSTargetMachine@llvm@@UEAA_NAEAVPassManagerBase@legacy@2@AEAVformatted_raw_ostream@2@W4CodeGenFileType@TargetMachine@2@_NPEBX4@Z)`.

kripken:
That error should have been fixed. Is that still a problem on latest incoming for you?

juj:
Indeed, works in incoming, and all looks good also after the merge.

PR #86, opened 2015-04-16
Add back a space that was removed in emscripten, is in LLVM.

jfbastien:
This just makes our diff smaller.


jfbastien:
Oops, wanted to base this on incoming, not master. Closing.

PR #96, opened 2015-04-22
Disable atomics more cleanly

jfbastien:
Atomics rewriting to PNaCl's own intrinsics wasn't quite clean, and was missing cmpxchg. This patch reverts the change from RewriteAtomics and just disable the pass in PNaClABISimplify.

### DIFF
--- a/lib/Transforms/NaCl/PNaClABISimplify.cpp
+++ b/lib/Transforms/NaCl/PNaClABISimplify.cpp
@@ -187,7 +187,8 @@ void llvm::PNaClABISimplifyAddPostOptPasses(Triple *T, PassManagerBase &PM) {
   if (!isEmscripten) // Handled by JSBackend.
     PM.add(createExpandGetElementPtrPass());
   // Rewrite atomic and volatile instructions with intrinsic calls.
-  PM.add(createRewriteAtomicsPass());
+  if (!isEmscripten) // No atomics currently.
+    PM.add(createRewriteAtomicsPass());
   // Remove ``asm("":::"memory")``. This must occur after rewriting
   // atomics: a ``fence seq_cst`` surrounded by ``asm("":::"memory")``
   // has special meaning and is translated differently.
--- a/lib/Transforms/NaCl/RewriteAtomics.cpp
+++ b/lib/Transforms/NaCl/RewriteAtomics.cpp
@@ -307,7 +307,6 @@ void AtomicVisitor::replaceInstructionWithIntrinsicCall(
 /// becomes:
 ///   %res = call T @llvm.nacl.atomic.load.i<size>(%ptr, memory_order)
 void AtomicVisitor::visitLoadInst(LoadInst &I) {
-  return; // XXX EMSCRIPTEN
   if (I.isSimple())
     return;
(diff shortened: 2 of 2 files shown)


kripken:
This removes 4 `return;`s, but there is a 5th case there: `visitAtomicCmpXchgInst`, which we actually do need. Could we check `isEmscripten` in those 4?

jfbastien:
@kripken OK I thought the lack of `return` on cmpxchg was a mistake! I'm not sure I understand how it's needed (and why not the others), is there a repro/test that we could tie this to?

juj:
For the Emscripten pthreads branch, I had to also add a `return;` to `visitAtomicCmpXchgInst` (https://github.com/juj/emscripten-fastcomp/commit/89ea50b6ab421bf26bd9789c0515ac7673d69793#diff-6dcdc7a6e10f4a3c4d284a86894ca0a3R385), or otherwise calling e.g. GCC cmpxchg intrinsic would cause a NaCl intrinsic to be generated, on which JSBackend would later on abort when it saw that as an unrecognized instruction (test at https://github.com/juj/emscripten/blob/pthreads/tests/pthread/test_pthread_gcc_atomics.cpp#L34 showed this iirc). I'm under the impression that we'd never need `visitAtomicCmpXchgInst` to execute(?).

jfbastien:
I think the right solution is to keep my patch as-is, but add handling of cmpxchg in JSBackend: handle the instruction instead of the PNaCl intrinsic. @juj's branch can then handle cmpxchg in a way that doesn't break threads :-)

I'll send an update later today.

kripken:
cc @sunfishcode , I'm not sure why we use the cmpxchg lowering pass.

sunfishcode:
I agree with @jfbastien. JSBackend should just handle cmpxchg directly.

jfbastien:
OK I'll do that.

jfbastien:
Hmm, I now remember why this was a PITA: `cmpxchg` returns a `{ T, i1 }` so it has `extractelement` and such.

@juj's [branch][1] has proper handling for these. Is his approach generally the one that's desired? I'd rather avoid reimplementing something that's already in that branch if I can  :-)

  [1]: https://github.com/juj/emscripten-fastcomp/blob/pthreads/lib/Target/JSBackend/JSBackend.cpp

juj:
@kripken and @sunfishcode had some comments about the handling there, and I don't yet know how to address that the best, it's been in my pondering list. At least it will need to detect when the `extractelement` comes as a result of the `cmpxchg` instruction and not try to process all extractelements blindly. Otherwise I imagine the scheme with `x$0` and `x$1` is safe for JS purposes at least.

kripken:
I think the issue there is that just appending `$0` or `$1` means that other parts of the backend aren't aware that there are really 2 variables there, etc. It's safer to do the transformation on LLVM IR.

jfbastien:
I'll close this for now, and leave the implementation as-is. It's probably worth cleaning up in a different way, either with @juj's approach or something else (teaching emscripten to lower all PNaCl atomic intrinsics could work, until we get full atomic support).

PR #97, opened 2015-05-02
Emit debug info only when the line number is greater than 0.

weitjong:
This PR is a bug fix for issue https://github.com/kripken/emscripten/issues/3401.


kripken:
Thanks!

PR #99, opened 2015-05-14
Add support for llvm_fabs_f32 in JSBackend.

juj:


lib/Target/JSBackend/CallHandlers.h
```diff
@@ -590,6 +590,7 @@ DEF_BUILTIN_HANDLER(sqrtl, Math_sqrt);
 DEF_BUILTIN_HANDLER(fabs, Math_abs);
 DEF_BUILTIN_HANDLER(fabsf, Math_abs);
 DEF_BUILTIN_HANDLER(fabsl, Math_abs);
+DEF_BUILTIN_HANDLER(llvm_fabs_f32, Math_abs);
```
kripken: I think this is probably right, but are you certain? `Math_abs` works on double precision, but here we need float precision. I assume this is fine in one sense because the input and output values are both going to be floats. However, I think it will break asm validation. Did you verify `asm2f` on the new test?

juj: Yeah, the test passes in all suites:

    test_llvm_fabs (test_core.asm1) ... ok
    test_llvm_fabs (test_core.asm2) ... ok
    test_llvm_fabs (test_core.asm3) ... ok
    test_llvm_fabs (test_core.asm2f) ... ok
    test_llvm_fabs (test_core.asm2g) ... ok
    test_llvm_fabs (test_core.asm1i) ... ok
    test_llvm_fabs (test_core.asm3i) ... ok
    test_llvm_fabs (test_core.asm2nn) ... ok

Manually building the test with `emcc tests/core/test_llvm_fabs.c -o a.html -s PRECISE_F32=1 -O1` does validate correctly for me. It does look things should be good here?

kripken: Sounds good then.

PR #102, opened 2015-06-16
Expand big switches

kripken:
For #3534

The 2048 number here is just a guess. We should find a number that helps avoid "too much recursion" and other browser parser errors, as mentioned in #3534.


kripken:
Updated with the 1024 limit and removing the old de-switch decision code, and merged.

PR #103, opened 2015-06-17
Add support for building SSE2 code to JS.

juj:


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -797,6 +803,22 @@ std::string JSWriter::getAssignIfNeeded(const Value *V) {
...
 
+std::string SIMDType(VectorType *t, bool caps = false) {
+  bool isInt = t->getElementType()->isIntegerTy();
+  int primSize = t->getElementType()->getPrimitiveSizeInBits();
+  int numElems = t->getNumElements();
+
+  return (caps ? (isInt ? "Int" : "Float") : (isInt ? "int" : "float")) + std::to_string(primSize) + 'x' + std::to_string(numElems);
+}
+
+std::string WrapMathFRound(const std::string &value, bool wrap) {
```
kripken: this is similar to `ensureFloat`. First of all, the name should begin with a lowercase letter. Second, how about also calling it `ensureFloat`, but overloaded with a second parameter which is a bool? And implementing it right after `ensureFloat`?

kripken:
Aside from one comment above, this looks good. I am not familiar with the SIMD correctness wrt the spec, though, perhaps @sunfishcode wants to review that aspect.

juj:
Thanks, renamed the function. Also did other changes to help SSE2 code generation in https://github.com/kripken/emscripten/pull/3542

juj:
Ping @sunfishcode , this is now ready to be reviewed and merged from my end. Merging this should be done at the same time with https://github.com/kripken/emscripten/pull/3542, since they go hand in hand.

PR #104, opened 2015-06-26
fix_recursive_mergelists_stack_smash

juj:
Avoids a recursive algorithm for merging use lists, which blows up the call stack and crashes LLVM parsing large .bc files on Windows.

Need to land this immediately/directly to fastcomp, since several partners are blocked on this right now. @sunfishcode , would you like to rally this upstream?

### DIFF
--- a/include/llvm/IR/Value.h
+++ b/include/llvm/IR/Value.h
@@ -502,16 +502,39 @@ private:
   /// \note Completely ignores \a Use::Prev (doesn't read, doesn't update).
   template <class Compare>
   static Use *mergeUseLists(Use *L, Use *R, Compare Cmp) {
-    Use *Merged;
-    mergeUseListsImpl(L, R, &Merged, Cmp);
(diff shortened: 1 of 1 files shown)


kripken:
Hopefully @sunfishcode can review this too.

@juj, do you know why this is a problem only on windows?

juj:
It was a difference in thread stack sizes. Actually, in the opencv.js test
case, @sunfishcode was able to crash the stack on his Linux as well, so
guess that windows just has a smaller default stack sizes.

kripken:
thanks, got it.

meanwhile lgtm, except please add an `XXX EMSCRIPTEN` so this is very noticeable in the source as a hack we need to remove.

 (but let's wait on @sunfishcode before merging)

juj:
Sure, added a `XXX EMSCRIPTEN` comment.

kripken:
Well, since so many people are having this issue, let's merge this in for now. We can adjust things later.

PR #106, opened 2015-07-14
making js output consistent each time it compiles the same sources

pkholland:
The goal is to make it so that compiling with identical sources produces identical output js files each time you compile.  The strategy is to use an "ordered" version of std::map and std::set whose syntax is the same as map and set, but whose begin() -> end() iterate elements in the order they were added to the container (unlike map and set which iterate according to the comparison operator of T).  Compile times with these modifications are nearly identical to the existing version on the projects I tried it on.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -2577,11 +2577,14 @@ void JSWriter::printFunctionBody(const Function *F) {
           }
           BlocksToConditions[BB] = Condition + (!UseSwitch && BlocksToConditions[BB].size() > 0 ? " | " : "") + BlocksToConditions[BB];
         }
-        for (BlockCondMap::const_iterator I = BlocksToConditions.begin(), E = BlocksToConditions.end(); I != E; ++I) {
-          const BasicBlock *BB = I->first;
+        std::set<const BasicBlock *> alreadyProcessed;
+        for (SwitchInst::ConstCaseIt i = SI->case_begin(), e = SI->case_end(); i != e; ++i) {
+          const BasicBlock *BB = i.getCaseSuccessor();
+          if (!alreadyProcessed.insert(BB).second)
+            continue;
           if (BB == DD) continue; // ok to eliminate this, default dest will get there anyhow
           std::string P = getPhiCode(&*BI, BB);
-          LLVMToRelooper[&*BI]->AddBranchTo(LLVMToRelooper[&*BB], I->second.c_str(), P.size() > 0 ? P.c_str() : NULL);
+          LLVMToRelooper[&*BI]->AddBranchTo(LLVMToRelooper[&*BB], BlocksToConditions[BB].c_str(), P.size() > 0 ? P.c_str() : NULL);
         }
         break;
       }
--- a/lib/Target/JSBackend/Relooper.h
+++ b/lib/Target/JSBackend/Relooper.h
@@ -44,8 +44,11 @@ struct Branch {
   void Render(Block *Target, bool SetLabel);
 };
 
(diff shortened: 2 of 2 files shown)


pkholland:
second commit is a minor improvement to the performance of the std::map case by avoiding an unnecessary, second lookup during operator[], and also an equally minor improvement in the naming of the two new classes to try to better suggest that begin() -> end() will be ordered according to how the elements were inserted.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -2577,11 +2577,14 @@ void JSWriter::printFunctionBody(const Function *F) {
           }
           BlocksToConditions[BB] = Condition + (!UseSwitch && BlocksToConditions[BB].size() > 0 ? " | " : "") + BlocksToConditions[BB];
         }
-        for (BlockCondMap::const_iterator I = BlocksToConditions.begin(), E = BlocksToConditions.end(); I != E; ++I) {
-          const BasicBlock *BB = I->first;
```
pkholland: This change causes the calls to AddBranchTo to be made in an order that is determined by the order of Si->case_begin() .. SI->case_end().  The original code makes the calls in the order of operator<(BasicBlock*, BasicBlock*).  That is, they were being made in an order determined by the address of the allocated BaskcBlock's.

kripken:
I ran tests and benchmarks. Overall the only issue I saw was a 5% slowdown on llc when building Unity, which ends up a 1% slowdown overall (llc is just one part).

I would be ok with a possible uniform slowdown like this, in return for deterministic builds. However, I worry the slowdown is not uniform, as `erase()` has gone from `O(log(N))` to `O(N)`, hasn't it? (And we do use `erase` in this codebase.)

pkholland:
Depends on which erase you are talking about, but in general I think the complexity has not changed in the way you are suggesting.  It has, however, changed in one case.

For InsertOrderedSet, the original code was only using (and my new class only provides)
   erase(const value_type val&)
That one is declared to have complexity O(log(N)) in std::set.  The new version does std::map::erase(const K& k) (O(log(N))) and std::list::erase(iterator position) which is constant time.

For InsertOrderedMap the original was using both erase(iterator position) erase(const Key& k).  The erase(const Key& k) is O(log(N)) for both the original and new code for the same reason as the InsertOrderedSet.

However,
The erase(iterator) case is amortized constant complexity in std::map, while my new version is O(log(N)) because I end up doing a find on the map, based on the key in the iterator.  Basically, I don't have a good way to go from a list::iterator back to the map::iterator without going through a map search.  I believe that the single place that this is used is:

https://github.com/kripken/emscripten-fastcomp/blob/incoming/lib/Target/JSBackend/Relooper.cpp#L973

I don't have a good way of knowing how often that line executes.  It's possible that the performance slow down is due to this one change, but I suspect it is more likely to be due to the additional memory allocations, etc... involved in maintaining both a map and set.  If it is this general memory issue then I think it would tend to be a fairly consistent slow down for all uses.

pkholland:
Also, note that I did slightly improve the performance of the Map::operator[] case, so if you 5% slowdown timing data was from prior to this change it might come out slightly better after the change.

kripken:
I did check with both commits. But that all makes sense, so I think this is good to merge.

pkholland:
One more.  The change in JSBackend.cpp does go from constant time to O(Log(N)) in the way it accesses the Block* that it passes to AddBranchTo.  But this only happens for N = the number of case statements in a single C switch statement, and I would think it is unusual for there to be very large numbers of this in realistic C code, but it is certainly possible.

kripken:
There are some switches with many thousands of elements that I've seen. But I am not that worried about logarithmic time, seems ok.

PR #107, opened 2015-07-20
Add ministring.h from kripken/Relooper

regular:
When compiling Relooper.cpp with emcc, the file `ministring.h` is missing. I copied it from kripken/Relooper/master.

    curl https://raw.githubusercontent.com/kripken/Relooper/master/ministring.h > lib/Target/JSBackend/ministring.h

### DIFF
--- a/lib/Target/JSBackend/ministring.h
+++ b/lib/Target/JSBackend/ministring.h
@@ -1,0 +1,35 @@
+
+// Tiny implementation of strings. Avoids linking in all of std::string
+
+#include <stdlib.h>
+#include <string.h>
+
+class ministring {
+  int used;
+  char *buffer;
+  int bufferSize;
+public:
+  ministring() : used(0), buffer(NULL), bufferSize(0) {}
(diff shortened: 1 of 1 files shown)


kripken:
Is this for building all of LLVM with emscripten? If not, then standalone builds of the Relooper, if people want that, I should probably do a better job of maintaining a standalone repo for it...

regular:
yes, it's for building all of llvm tools with emscripten. You think you can merge this?

kripken:
I see. Ok, let's get that file in there, then. I assume you didn't modify it, it's the original `ministring.h`? Then to keep the git authors correct in the history, I can just push it myself, if that's ok with you?

kripken:
Actually, on second thought, it would be better to remove the `ifdef`s and not use `ministring.h`. For a project the size of LLVM, might as well just use the normal string code since it'll be linked anyhow. Also better in terms of code simplicity, to remove any differences between the build types.

regular:
Agreed. I'll send another PR then.

PR #108, opened 2015-07-22
Replace ministring with std::string in emcc build of Relooper.cpp

regular:
fixes broken build due to missing ministring.h

### DIFF
--- a/lib/Target/JSBackend/Relooper.cpp
+++ b/lib/Target/JSBackend/Relooper.cpp
@@ -9,13 +9,7 @@
 #include <stdlib.h>
 #include <list>
 #include <stack>
-
-#if EMSCRIPTEN
-#include "ministring.h"
-#else
 #include <string>
-typedef std::string ministring;
-#endif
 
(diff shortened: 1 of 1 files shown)


kripken:
Thanks!

regular:
Thank you too!
In case you are interested: [this](http://regular.github.io/fastcomp.js/demo.html) is what I use the emcc build of fastcomp for. (the layout might look familiar to you :) )
And [here](http://github.com/regular/make-llvm-bitcode-for-javascript) is the script that automates all of the BitCode generation. (I can get rid of that `curl` line now ...)

kripken:
Nice!

One note, when I wrote some bad LLVM IR and tried to compile it, it replaced the IR with the error message, so the IR was lost, which surprised me.

regular:
Uh, that's not intentionally, just a side-effect of not having proper error handling yet. I look into it. 
I plan to have three areas: source assembler/optimized assembler/js
Thanks for the feedback! regular/fastcomp.js#5

PR #115, opened 2015-09-03
Fix vs2013 build

juj:
Fixes Visual Studio 2013 incompatibilities that we got from upstream.

CC @jfbastien .

### DIFF
--- a/include/llvm/Bitcode/NaCl/NaClRandNumGen.h
+++ b/include/llvm/Bitcode/NaCl/NaClRandNumGen.h
@@ -20,6 +20,7 @@
 
 #include "llvm/ADT/StringRef.h"
 #include <random>
+#include <stdint.h>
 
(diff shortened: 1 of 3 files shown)


jfbastien:
Thanks for pointing this out :-)

gouletr:
Great! We can close https://github.com/kripken/emscripten/issues/3738 when this is merged.

kripken:
@jfbastien , do these look good? If you don't have a quick answer, then @juj lgtm but with `XXX EMSCRIPTEN` comments.

In other words lgtm either way, but with comments if these are things pnacl isn't sure they're going to do.

jfbastien:
lgtm, I asked kschimpf to apply the same patch here:
https://codereview.chromium.org/1306183006/

kripken:
Great, thanks.

PR #116, opened 2015-09-03
Fix NaNs in SIMD

juj:
When encountering float literals that have noncanonical NaN bit patterns in a SIMD vector, don't attempt to create a float SIMD vector directly out of them, since JS source can't represent NaNs with noncanonical bits. Instead, create an int SIMD vector and cast it to float one to detour to the SIMD vector with correct bits intact. Fixes https://github.com/kripken/emscripten/issues/2840,  https://github.com/kripken/emscripten/issues/3560, https://github.com/kripken/emscripten/issues/3010 and https://github.com/kripken/emscripten/issues/3403.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -126,7 +126,7 @@ namespace {
   #define ASM_FFI_IN 4 // FFI return values are limited to things that work in ffis
   #define ASM_FFI_OUT 8 // params to FFIs are limited to things that work in ffis
   #define ASM_MUST_CAST 16 // this value must be explicitly cast (or be an integer constant)
-  #define ASM_FORCE_FLOAT_AS_INTBITS 32 // if the value is a float, it should be returned as an integer representing the float bits (or NaN canonicalization will eat them away)
+  #define ASM_FORCE_FLOAT_AS_INTBITS 32 // if the value is a float, it should be returned as an integer representing the float bits (or NaN canonicalization will eat them away). This flag cannot be used with ASM_UNSIGNED set.
   typedef unsigned AsmCast;
 
   typedef std::map<const Value*,std::string> ValueMap;
@@ -1303,18 +1303,9 @@ std::string JSWriter::getConstant(const Constant* CV, AsmCast sign) {
     } else {
       const APFloat &flt = CFP->getValueAPF();
       APInt i = flt.bitcastToAPInt();
-      if ((sign & ASM_UNSIGNED)) {
-        if (i.getBitWidth() <= 64 && *i.getRawData() < 1000000) {
-          // Small ints as digits, where adding '0x' makes the string longer. "1000000" == "0xF4240" is equal length.
-          return utostr(*i.getRawData());
-        } else {
-          // Large ints as hex.
-          return std::string("0x") + utohexstr(*i.getRawData());
-        }
-      } else {
-        if (i.getBitWidth() == 32) return itostr((int)(uint32_t)*i.getRawData());
-        else return itostr(*i.getRawData());
-      }
+      assert(!(sign & ASM_UNSIGNED));
+      if (i.getBitWidth() == 32) return itostr((int)(uint32_t)*i.getRawData());
+      else return itostr(*i.getRawData());
     }
   } else if (const ConstantInt *CI = dyn_cast<ConstantInt>(CV)) {
     if (sign != ASM_UNSIGNED && CI->getValue().getBitWidth() == 1) {
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -121,6 +121,7 @@ namespace {
   #define ASM_FFI_IN 4 // FFI return values are limited to things that work in ffis
   #define ASM_FFI_OUT 8 // params to FFIs are limited to things that work in ffis
   #define ASM_MUST_CAST 16 // this value must be explicitly cast (or be an integer constant)
+  #define ASM_FORCE_FLOAT_AS_INTBITS 32 // if the value is a float, it should be returned as an integer representing the float bits (or NaN canonicalization will eat them away)
   typedef unsigned AsmCast;
 
   typedef std::map<const Value*,std::string> ValueMap;
@@ -495,7 +496,15 @@ namespace {
 
       // Emscripten has its own spellings for infinity and NaN.
       if (flt.getCategory() == APFloat::fcInfinity) return ensureCast(flt.isNegative() ? "-inf" : "inf", CFP->getType(), sign);
-      else if (flt.getCategory() == APFloat::fcNaN) return ensureCast("nan", CFP->getType(), sign);
+      else if (flt.getCategory() == APFloat::fcNaN) {
+        APInt i = flt.bitcastToAPInt();
+        if ((i.getBitWidth() == 32 && i != APInt(32, 0x7FC00000)) || (i.getBitWidth() == 64 && i != APInt(64, 0x7FC0000000000000ULL))) {
+          // If we reach here, things have already gone bad, and JS engine NaN canonicalization will kill the bits in the float. However can't make
+          // this a build error in order to not break people's existing code, so issue a warning instead.
+          prettyWarning() << "Warning: ftostr() cannot represent a NaN literal '" << CFP << "' with custom bit pattern without erasing bits!\n";
+        }
+        return ensureCast("nan", CFP->getType(), sign);
+      }
 
       // Request 9 or 17 digits, aka FLT_DECIMAL_DIG or DBL_DECIMAL_DIG (our
       // long double is the the same as our double), to avoid rounding errors.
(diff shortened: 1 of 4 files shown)


kripken:
I have no objections, leaving to @sunfishcode since I'm not sure what the big picture for SIMD canonicalization is.

kripken:
Is there a testcase for this, to help debug?

juj:
See https://github.com/kripken/emscripten/issues/2840 and https://github.com/kripken/emscripten/issues/3560.

kripken:
Oh right, sorry - I missed those.

juj:
Ok, updated the pull request, this is good for review now. Tested to pass the `asm1` suite.

sunfishcode:
(Reposting my comment from https://github.com/kripken/emscripten/issues/3560)

After talking with @kripken and looking at the testcases, it looks like NaN bit patterns in scalar values are a lost cause in JS and therefore also asm.js, however NaN bit patterns in SIMD values are doable. I wrote up this patch:

sunfishcode/emscripten-fastcomp@7644927

which implements a fix just for vector constants, in a very simple way. I plan to clean this code up a bit, and it's not really tested yet, but I'm posting it here to see what people think. Is this a feasible way forward?

juj:
I am a bit confused. The first commit in my pull request already does this (plus it removes the code duplication we have in `getConstantVector(ConstantVector)` vs `getConstantVector(ConstantDataVector)`). Alon asked me to add the second commit https://github.com/juj/emscripten-fastcomp/commit/01e549e0afd1d58d1845440d9fb37f644df4e696 to fix the printing for non-vector case as well so that we are not broken with nan printing anywhere. If that is not desirable after deliberation, then we should either issue a warning or an error, but never silently print wrong value as a nan with bits discarded at compile time (the way that my first commit in https://github.com/juj/emscripten-fastcomp/commit/0f9d3d1038e2ec398e5e8741d85aadd9d3ee39e4#diff-7212822d603be752dbb5c5d4619a359eR504 did before the change).

I do see your code is cleaner in some respects (`VectorType::getInteger(C->getType());` is very nice), it would be nicer that cleanup be refactored on top of my pull request instead of a separate implementation which immediately conflicts with the pending pull request.

In either case, I'm ok with whichever code, as long as the test cases that I pointed out in the PR still work, and that https://github.com/kripken/emscripten/pull/3740 also works afterwards, and that there is no possibility of the compiler silently emitting nans with wrong bits at compiletime (The implementation of `ftostr` is currently incorrect).

kripken:
The second commit can't work though, can it? AFAICT, we have no way to fix the non-vector case. Scalar JS variables simply cannot hold non-canonicalized nans, so that's a fight we can't win. I spent some time investigating this with @sunfishcode today.

We can fix the vector-receiving-a-constant case, though (but not a variable), which it looks like both your first patch and @sunfishcode's do. Do they do the same thing, though? Yours seems to handle more cases in more places in the code, but I don't quite understand yet in what ways.

@sunfishcode, perhaps you can test your patch on the other 3 cases in the list on top, to get an idea if the coverage is the same?

juj:
In my commit, I try to print a vector as `nan` only if the bits in nan are non-canonical, but still print as `nan` if they are canonical. I try to find the smallest length integer representation (hex vs integral), those can be discarded for simplicity, they are rather small optimizations if not desirable. Also, @sunfishcode 's commit doesn't currently handle splats with non-canonical nans.

I did see when building MathGeoLib without the second commit `ftostr` patch, that compilation did print warning messages about the compiler canonicalizing bits on the code during optimizations, even though MathGeoLib doesn't utilize non-canonical NaN scalar literals in a float anywhere since https://github.com/juj/MathGeoLib/commit/d97867d858f5947709f16079cbaa9cebedca1bed .

juj:
Ideally I think it would be nice to have a compiler flag on how the compiler should treat noncanonical nans it sees (warning, error, or route through memory).

kripken:
I think there isn't a standardized canonical nan, implementations [must only be internally consistent](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-setvalueinbuffer) (items 9&10), so I think we have to always emit a bitcast here.

For the same reason, I'm not sure how we can warn on this.

juj:
Let's be pragmatic, we aren't developing against unknown browsers that might exist some day. Both Firefox and Safari canonicalize to `0x7fc00000` and `0x7fc00000 00000000`. If someone comes up with a browser that canonicalizes to something else, we can reconsider. I really like the approach of making it a compiler option of 1) error, 2) warning+route through memory, 3) no warning+route through memory. 

If that is not acceptable, then we should emit a list of nan bits that were generated at compile time, and at startup check that list and issue a warning/error of bits having been lost. Or simply check against the single expected form of NaN at startup.

I dont want to accept this kind of potential source of silent failure that `ftostr` currently has, especially when LLVM optimizer is in play, which freely moves scalar vs vector loads and stores like this, and some NaN issues only occur when LLVM optimizations are applied. We have all the information at compile-time, and all the information at browser startup time. Being broken at compile time is a much better option than being silently broken during execution.

In my mind, "don't canonicalize NaNs" is a feature just like SharedArrayBuffer, SIMD and the rest, whatever we make the compiler emit, we could check at startup and detect if the runtime doesn't match with what is expected. In the long run, NaN canonicalization hurts performance and we should advocate that implementations should not do it anyways when running high performance asm.js code, and by making our compiler support such a target might lead Safari/jsc to reconsider as well.

sunfishcode:
@juj Ah, sorry, I missed the history here. I'm ok if we proceed with your patch, with cleanups including `VectorType::getInteger(C->getType())` etc. being integrated into it.

I don't currently think we need to preserve NaN bit patterns in *scalar* literals in asm.js, since scalar NaNs aren't going to preserve their bits no matter what we do, because of JS semantics. Fortunately, as far as I can tell, the testcases here are all SIMD-specific idioms: creating lane masks and `xmmintrin.h` having bitwise and/or operate on floating-point types. Consequently, I would prefer to keep Emscripten simpler and avoid the complexity of preserving NaN bits for scalar literals. A warning when we silently canonicalize a literal is reasonable though.

kripken:
If we want to consider only existing VMs, then what about IE/Edge, Nashorn, Rhino? Does they canonicalize, and if so, to what?

juj:
Neither IE or Edge canonicalize. I don't know about the others.

sunfishcode:
For the purposes of issuing a warning on scalar NaN literals that we can't encode, I claim it's reasonable to assume 0x7ff8000000000000 is the canonical pattern for 64-bit (@juj I believe your pattern above is a typo). I don't think we need to worry about what every platform does here; this should just be a best-effort warning anyway.

kripken:
Fair enough. Let's just mention that in the warning.

juj:
Both Firefox and Safari canonicalize to the values `0x7fc00000` and `0x7fc00000 00000000` in my tests so I don't think it's a typo.

Updated the pull request to revert the scalar casting and replaced it with a warning, and changed to use `VectorTyoe::getInteger()`. This is ready to go.

sunfishcode:
@juj 0x7fc0000000000000 isn't a NaN bit pattern.

This code:

```
var a = new ArrayBuffer(8);
var i = new Int32Array(a);
var d = new Float64Array(a);

d[0] = NaN;
console.log(i[1].toString(16));
console.log(i[0].toString(16));
```
prints this in SpiderMonkey:
```
7ff80000
0
```

juj:
Oh thanks! Indeed, I stand corrected, my code to test was buggy. Updated the pull request.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1276,11 +1291,28 @@ std::string JSWriter::getConstant(const Constant* CV, AsmCast sign) {
...
+          // Small ints as digits, where adding '0x' makes the string longer. "1000000" == "0xF4240" is equal length.
+          return utostr(*i.getRawData());
+        } else {
+          // Large ints as hex.
+          return std::string("0x") + utohexstr(*i.getRawData());
+        }
+      } else {
+        if (i.getBitWidth() == 32) return itostr((int)(uint32_t)*i.getRawData());
+        else return itostr(*i.getRawData());
+      }
```
sunfishcode: Do we need separate code for signed and unsigned cases here? This is only used to print bits which should be floating-point, so it's not clear why we should have multiple cases.

juj: I don't know. I followed the implementation according to the comment of what I understood `ASM_UNSIGNED` to mean, i.e. the that is specified, the literal is being asked to be printed out as unsigned. However if that is not needed, then I can remove the unsigned case?

juj:
Ping @sunfishcode ? Could you check this again, I saw you merged the other pull requests.

juj:
Rebased to latest incoming, and added a commit that fixes https://github.com/kripken/emscripten/issues/3784. This is ok to review again @sunfishcode ?

lib/Target/JSBackend/JSBackend.cpp
sunfishcode: This code is only used when ASM_FORCE_FLOAT_AS_INTBITS is used, so it doesn't need to be able to do both signed and unsigned, because its user is just going to bitcast it to a float. I think you can do just the signed case and add an assert that ASM_UNSIGNED is not set.

juj: Ok, enforced that `ASM_UNSIGNED` cannot be passed when `ASM_FORCE_FLOAT_AS_INTBITS` is specified.

juj:
Thanks for the review, all comments should now be addressed. How does this look?

sunfishcode:
Thanks!

PR #117, opened 2015-09-04
Float64x2 working in simd polyfill

juj:
Fixes codegen so that Float64x2 code works equally in node.js and SpiderMonkey.

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -807,12 +807,45 @@ DEF_BUILTIN_HANDLER(emscripten_float64x2_fromInt32x4Bits, SIMD_Float64x2_fromInt
 DEF_BUILTIN_HANDLER(emscripten_float64x2_fromInt32x4, SIMD_Float64x2_fromInt32x4);
 DEF_BUILTIN_HANDLER(emscripten_float64x2_fromFloat32x4Bits, SIMD_Float64x2_fromFloat32x4Bits);
 DEF_BUILTIN_HANDLER(emscripten_float64x2_fromFloat32x4, SIMD_Float64x2_fromFloat32x4);
+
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/CallHandlers.h
```diff
@@ -807,12 +807,45 @@ DEF_BUILTIN_HANDLER(emscripten_float64x2_fromInt32x4Bits, SIMD_Float64x2_fromInt
 DEF_BUILTIN_HANDLER(emscripten_float64x2_fromInt32x4, SIMD_Float64x2_fromInt32x4);
 DEF_BUILTIN_HANDLER(emscripten_float64x2_fromFloat32x4Bits, SIMD_Float64x2_fromFloat32x4Bits);
 DEF_BUILTIN_HANDLER(emscripten_float64x2_fromFloat32x4, SIMD_Float64x2_fromFloat32x4);
+
+static std::string castBool64x2ToInt32x4(const std::string &valueStr) {
+  return std::string("SIMD_Int32x4_fromBool64x2Bits(") + valueStr + ')';
```
sunfishcode: The SIMD.js spec doesn't have fromBool*Bits functions. As far as I know, the only way to do this is to use a select.

juj: This function is something I added myself to emulate, since I didn't see another way as there is no such select two -> four expanding type of code. See https://github.com/juj/emscripten/commit/15705a5138849005845ae59c381f9539ce629c28 .

juj:
Actually, let me leave the shift rename commit out from here now, since that breaks validation. We should apply it separately when https://bugzilla.mozilla.org/show_bug.cgi?id=1201934 occurs.

juj:
Updated this PR to omit the shift rename. This is good to go.

sunfishcode:
lgtm

PR #120, opened 2015-09-22
Allow multiple arities per EM_ASM block, fixes kripken/emscripten#3804.

jjrv:
PR for kripken/emscripten#3804.
Make a set of different arities in template expansions of each `EM_ASM` block.
They need to be ordered to take a shortcut in `emscripten.py` so there's a new typedef `OrderedIntSet` in case `IntSet` is later changed to `std::unordered_set`.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -126,6 +126,7 @@ namespace {
   typedef std::map<const Value*,std::string> ValueMap;
   typedef std::set<std::string> NameSet;
   typedef std::set<int> IntSet;
+  typedef std::set<int> OrderedIntSet;
   typedef std::vector<unsigned char> HeapData;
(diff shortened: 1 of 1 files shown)


kripken:
Looks perfect, just needs to wait on the parallel https://github.com/kripken/emscripten/pull/3806 to land together.

jjrv:
A small change in the other PR lifted the ordered set requirement here, so I renamed that variable and removed a now unused one.

jjrv:
Maybe the structure of `asmConstArities` and the code emitting it is now complicated enough to benefit from a clarifying comment.

kripken:
Merged, thanks!

Meanwhile I was thinking that the data structures could be simpler here if we just outputted the max for each id, and a list of all the arities seen overall. However, there seems little downside to providing the full information as this pull does, perhaps we will need it later.

PR #121, opened 2015-09-23
Fix compilation issue : “template specialization requires 'template<>'"

gogoprog:
I encountered this issue while compiling clang-incoming-64bit on Linux


gogoprog:
Fixed during my PR ;)

PR #122, opened 2015-09-24
Say that bad SIMD types are completely unacceptable

kripken:
This adds some `TargetTransformInfo` logic to deny some bad SIMD types, seen when autovectorizing Bullet.

Btw, one of the things it tries is to vectorize things like `<4 x btVector3*>`, which I suppose we could support in theory, as it's just a 32-bit int.

### DIFF
--- a/lib/Target/JSBackend/JSTargetTransformInfo.cpp
+++ b/lib/Target/JSBackend/JSTargetTransformInfo.cpp
@@ -21,6 +21,7 @@
 #include "llvm/CodeGen/BasicTTIImpl.h"
 #include "llvm/IR/Module.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
 #include "llvm/Target/CostTable.h"
 #include "llvm/Target/TargetLowering.h"
 using namespace llvm;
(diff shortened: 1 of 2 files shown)


kripken:
cc @juj, I saw this with `--llvm-lto 2 -s SIMD=1` on Bullet. I think you were looking at Bullet too?

kripken:
This is probably especially important given https://github.com/kripken/emscripten/commit/3d5ec1e1b737404331366687ed6bfdd85e39d582 which removed us forcing the vector width to 4. Although I don't actually see downsides to doing that in the test suite, but after this pull we should once more have a guarantee of not getting `!= 4` vector lengths.

lib/Target/JSBackend/JSTargetTransformInfo.cpp
```diff
@@ -42,30 +43,31 @@ unsigned JSTTIImpl::getRegisterBitWidth(bool Vector) {
   return 32;
 }
 
+static const unsigned Nope = 65536;
+
+// Certain types are fine, but some vector types must be avoided at all Costs.
+static bool isOkType(Type *Ty) {
+  if (VectorType *VTy = dyn_cast<VectorType>(Ty)) {
+    if (VTy->getNumElements() != 4 || !(VTy->getElementType()->isIntegerTy(1) ||
```
juj: What does it mean for a type to be ok? If a type is ok if it is a supported type that the JSBackend codegen is able to turn into functional SIMD code, then the good types are `Int8x16`, `Int16x8`, `Int32x4` (and unsigned variants), `Float32x4` and `Float64x2` (which is not yet officially in SIMD.js, but critical for us, so it was retained in the polyfill and Firefox also supports it, but doesn't yet validate). Additionally, we fake certain LLVM types in two ways, which makes more types ok:
    - single bit types are treated as being full 128-bit width: `Int1x16` -> `Int8x16`, `Int1x8` -> `Int16x8`, `Int1x4` -> `Int32x4`
    - any LLVM type with smaller number of lanes is expanded to enough lanes to be full 128-bit width: e.g. `Int8x5`-> `Int8x16` and `Int32x3` -> `Int32x4` and so on.

Both of these kind of fakings are done when we emit the JS code strings, so LLVM doesn't even know it's being done.

Bad types are `Int64x2` and `Int128x1`, which we have no clue what to do with (we could stuff the bits to to e.g. `Int32x4`, but if one happens to do arithmetic with them later on, we are burned). Also single bit types that are neither full 128.bit wide are ambiguous, I'm not sure what will happen currently if those are encountered, e.g. `Int1x2` (would it become `Int8x16` or `Int32x4`? It's possible that this would just work, when we do one widening first and then the second one, favoring one over the other).

So perhaps a good type is a
   a) a 32-bit float type with <= 4 lanes, or
   b) a 64-bit float type with <= 2 lanes, or
   c) an integer type with total number of bits <= 128 and lane width <= 32 ?

juj:
Looks good, this probably fixes issues with our autovectorization creating bad new int types, 64-bit and 128-bit specifically? Perhaps the decision to which types to allow is overly strict, and a more loose one could be used (see above comment), but not sure if that would win much.

kripken:
I'm curious about `<2 x float32>` for example - I understand we add zero lanes, as you say? But I wonder if it doesn't remove any effective speedup when doing so. This pull just affects the autovectorizer, so it seems safer to tell it to avoid the suboptimal cases. Or is it common natively as well to add zero lanes, and it's still worth it?

kripken:
In any case I'd like to land this as-is, to avoid risks from https://github.com/kripken/emscripten/commit/3d5ec1e1b737404331366687ed6bfdd85e39d582 as mentioned above. We can investigate allowing further autovectorization later, as an optimization followup.

sunfishcode:
lgtm; I agree that investigating further autovectorization later makes sense.

PR #123, opened 2015-10-03
Handle bitcast of <4 x i32> to i128

kripken:
Possible fix for https://github.com/kripken/emscripten/issues/3789

GVN in the optimizer emits `  %2 = bitcast <4 x i32> %a to i128  ` to avoid a load from memory. There doesn't seem to be a way to tell GVN not do that.

This change adds support to lower that bitcast, by extracting the 4 elements into two i64s, which would then also be broken down later into legal i32s. This is probably not that fast in general, unless the optimizer is run and also manages to do something useful.

### DIFF
--- a/lib/Transforms/NaCl/ExpandLargeIntegers.cpp
+++ b/lib/Transforms/NaCl/ExpandLargeIntegers.cpp
@@ -572,17 +572,26 @@ static void convertInstruction(Instruction *Inst, ConversionState &State,
     State.recordConverted(Select, {Lo, Hi});
 
   } else if (BitCastInst *BitCast = dyn_cast<BitCastInst>(Inst)) {
-    // XXX EMSCRIPTEN handle bitcast <4 x i32> to i128
+    // XXX EMSCRIPTEN handle bitcast <4 x i32|float> to i128
     Value *Input = BitCast->getOperand(0);
     assert(Input->getType()->isVectorTy());
     VectorType *VT = cast<VectorType>(Input->getType());
-    assert(VT->getNumElements() == 4 && VTy->getElementType()->isIntegerTy(32));
+    assert(VT->getNumElements() == 4);
+    Type *ET = VT->getElementType();
+    assert(ET->isIntegerTy(32) || ET->isFloatTy());
 
     Type *I32 = Type::getInt32Ty(VT->getContext());
     Value *P0 = IRB.CreateExtractElement(Input, ConstantInt::get(I32, 0), Twine(Name, ".p0"));
     Value *P1 = IRB.CreateExtractElement(Input, ConstantInt::get(I32, 1), Twine(Name, ".p1"));
     Value *P2 = IRB.CreateExtractElement(Input, ConstantInt::get(I32, 2), Twine(Name, ".p2"));
     Value *P3 = IRB.CreateExtractElement(Input, ConstantInt::get(I32, 3), Twine(Name, ".p3"));
+
+    if (ET->isFloatTy()) {
+      P0 = IRB.CreateBitCast(P0, I32, Twine(Name, ".i0"));
+      P1 = IRB.CreateBitCast(P1, I32, Twine(Name, ".i1"));
+      P2 = IRB.CreateBitCast(P2, I32, Twine(Name, ".i2"));
+      P3 = IRB.CreateBitCast(P3, I32, Twine(Name, ".i3"));
+    }
 
     Type *I64 = Type::getInt64Ty(VT->getContext());
     P0 = IRB.CreateZExt(P0, I64, Twine(Name, ".p0.64"));
--- a/lib/Transforms/NaCl/ExpandLargeIntegers.cpp
+++ b/lib/Transforms/NaCl/ExpandLargeIntegers.cpp
@@ -571,6 +571,29 @@ static void convertInstruction(Instruction *Inst, ConversionState &State,
     Value *Hi = IRB.CreateSelect(Cond, True.Hi, False.Hi, Twine(Name, ".hi"));
     State.recordConverted(Select, {Lo, Hi});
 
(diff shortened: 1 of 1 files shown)


lib/Transforms/NaCl/ExpandLargeIntegers.cpp
```diff
@@ -571,6 +571,38 @@ static void convertInstruction(Instruction *Inst, ConversionState &State,
     Value *Hi = IRB.CreateSelect(Cond, True.Hi, False.Hi, Twine(Name, ".hi"));
     State.recordConverted(Select, {Lo, Hi});
 
+  } else if (BitCastInst *BitCast = dyn_cast<BitCastInst>(Inst)) {
+    // XXX EMSCRIPTEN handle bitcast <4 x i32|float> to i128
+    Value *Input = BitCast->getOperand(0);
+    assert(Input->getType()->isVectorTy());
+    VectorType *VT = cast<VectorType>(Input->getType());
+    assert(VT->getNumElements() == 4);
+    Type *ET = VT->getElementType();
```
juj: I'm getting a crash that points to this line `Type *ET = VT->getElementType();`:

	0x00007FF7B2A11B82 (0x0000002D93F04458 0x0000002D854305E8 0x0000000000000000 0x0000002D93F03CE8), convertInstruction() + 0xE82 bytes(s), c:\code\emsdk\clang\fastcomp\src\lib\transforms\nacl\expandlargeintegers.cpp, line 580
	0x00007FF7B2A1614F (0x0000002DFEE4B988 0x0000002D86AC0050 0x0000002D84F418A0 0x0000002D93019D90), `anonymous namespace'::ExpandLargeIntegers::runOnFunction() +0x29F bytes(s), c:\code\emsdk\clang\fastcomp\src\lib\transforms\nacl\expandlargeintegers.cpp, line 637
	0x00007FF7B2802DB6 (0x0000002D84F418A0 0x0000002DFEE4B988 0x0000002DFEE4BA00 0x0000000000000001), llvm::FPPassManager::runOnFunction() + 0x156 bytes(s), c:\code\emsdk\clang\fastcomp\src\lib\ir\legacypassmanager.cpp, line 1538 + 0xF byte(s)
	0x00007FF7B2802F0F (0x0000000000000000 0x0000002D86AC7C01 0x0000002D86AC0050 0x0000002DFEE4BA98), llvm::FPPassManager::runOnModule() + 0x3F bytes(s), c:\code\emsdk\clang\fastcomp\src\lib\ir\legacypassmanager.cpp, line 1557
	0x00007FF7B2803119 (0x0000000000000013 0x0000000000000000 0x0000000000000003 0x0000002DFEE4B9C0), `anonymous namespace'::MPPassManager::runOnModule() + 0x1D9 bytes(s), c:\code\emsdk\clang\fastcomp\src\lib\ir\legacypassmanager.cpp, line 1616 + 0xF byte(s)
	0x00007FF7B2802901 (0x0000002D93888730 0x0000002DFED1F5E0 0x0000000000000000 0x0000002DFEE43E40), llvm::legacy::PassManagerImpl::run() + 0xD1 bytes(s), c:\code\emsdk\clang\fastcomp\src\lib\ir\legacypassmanager.cpp, line 1724
	0x00007FF7B22B9622 (0x00007FF7B3094020 0x0000000000000000 0x0000000000000000 0x00007FF7B3094020), main() + 0x16A2 bytes(s), c:\code\emsdk\clang\fastcomp\src\tools\opt\opt.cpp, line 792
	0x00007FF7B3093F73 (0x0000000000000000 0x0000000000000000 0x0000000000000000 0x0000000000000000), __tmainCRTStartup() + 0x10F bytes(s), f:\dd\vctools\crt\crtw32\dllstuff\crtexe.c, line 626 + 0x19 byte(s)
	0x00007FFEB99413D2 (0x00007FFEB99413B0 0x0000000000000000 0x0000000000000000 0x0000000000000000), BaseThreadInitThunk() + 0x22 bytes(s)
	0x00007FFEBBF75454 (0x0000000000000000 0x0000000000000000 0x0000000000000000 0x0000000000000000), RtlUserThreadStart() + 0x34 bytes(s)
	warning: incorrect target triple '' (did you use emcc/em++ on all source files and not clang directly?)

I think my build had asserts disabled (RelWithDebInfo).

juj: Doing a debug build now, which will take a while. Perhaps operand 0 might not always be a vector type(?)

juj: It looks like there is a bitcast `%42 = bitcast i128 %41 to <4 x i32>` in the file, on which the code above fails and in debug build the assert is hit.

sunfishcode:
The overall approach lgtm. It looks like from @juj's comments that you'll need to handle the opposite bitcast as well.

lib/Transforms/NaCl/ExpandLargeIntegers.cpp
kripken: Yes, it might not be a vector type, in which case we have another bug to handle.

kripken: On which codebase is the crash? Send me the bitcode if you want.

kripken:
Talking to @sunfishcode sunfish about those two bugs (this and https://github.com/kripken/emscripten/issues/3788 ), i think we both now lean towards just disabling those sroa and gvn features. we know that

1. Those passes never emitted those patterns before, or the compiler would have crashes, so they are super rare, and
2. We can't emit good code here. We must do bitcasts, and those are worse for us than for native. We should just to another load from memory, most likely, which is what disabling sroa would do.

If that sounds good, then I can look into hacking up those passes to avoid this issue. @sunfishcode suggested adding a warning which sounds good. The main risk here is hacking up core LLVM code, and future merges, of course.

kripken:
Looks like objections, so will open a pull for the other approach.

PR #124, opened 2015-10-06
Prevent LLVM's optimizer from emitting silly vector types

kripken:
Fixes https://github.com/kripken/emscripten/issues/3788 and https://github.com/kripken/emscripten/issues/3789 , by modifing GVN and SROA to not emit vector types, and so avoid costly bitcasts for us.

### DIFF
--- a/lib/Transforms/Scalar/GVN.cpp
+++ b/lib/Transforms/Scalar/GVN.cpp
@@ -1833,6 +1833,18 @@ bool GVN::processLoad(LoadInst *L) {
 
   // ... to a pointer that has been loaded from before...
   MemDepResult Dep = MD->getDependency(L);
+
+  // XXX EMSCRIPTEN: avoid GVNing between vector and non-vector types, as
+  //                 our bitcasts are very expensive
+  if (Dep.getInst()) {
+    if (Dep.getInst()->getType()->isVectorTy() ^ L->getType()->isVectorTy())
+      return false;
+    if (StoreInst *DepSI = dyn_cast<StoreInst>(Dep.getInst())) {
+      if (DepSI->getOperand(0)->getType()->isVectorTy() ^ L->getType()->isVectorTy())
+        return false;
+    }
+  }
+
   const DataLayout &DL = L->getModule()->getDataLayout();
 
   // If we have a clobber and target data is around, see if this is a clobber
--- a/lib/Transforms/Scalar/SROA.cpp
+++ b/lib/Transforms/Scalar/SROA.cpp
@@ -3996,8 +3996,8 @@ AllocaInst *SROA::rewritePartition(AllocaInst &AI, AllocaSlices &AS,
 
   bool IsIntegerPromotable = isIntegerWideningViable(P, SliceTy, DL);
 
-  VectorType *VecTy =
-      IsIntegerPromotable ? nullptr : isVectorPromotionViable(P, DL);
+  VectorType *VecTy = nullptr; // XXX EMSCRIPTEN - never promote to vectors
+      //IsIntegerPromotable ? nullptr : isVectorPromotionViable(P, DL); // waka
   if (VecTy)
     SliceTy = VecTy;
 


kripken:
@sunfishcode: These should be simple modifications. I tried to be conservative and just avoid vector <-> non-vector ops in GVN, so GVN will still operate on vectors <-> vector, which maybe helps in some cases.

We should run some vector/SIMD benchmarks on this, to be sure - @juj, do you have any?

juj:
@kripken : there is `python tests\benchmark_sse1.py` that is the only current benchmark. There's also the mandelbrot demo, although I wouldn't use that to benchmark anything, since it's not representative.

I'll try this with the Urho3D bullet codebase. Not sure how to do a comparative benchmark since it doesn't compile without.

kripken:
Ok, on the sse1 benchmark there is 0 code change (nice to have deterministic builds...).

lib/Transforms/Scalar/GVN.cpp
```diff
@@ -1833,6 +1833,18 @@ bool GVN::processLoad(LoadInst *L) {
 
   // ... to a pointer that has been loaded from before...
   MemDepResult Dep = MD->getDependency(L);
+
+  // XXX EMSCRIPTEN: avoid GVNing between vector and non-vector types, as
+  //                 our bitcasts are very expensive
+  if (Dep.getInst()) {
+    if (Dep.getInst()->getType()->isVectorTy() ^ L->getType()->isVectorTy())
```
sunfishcode: I suggest `!=` over `^` here and below, since these are single boolean values.

lib/Transforms/Scalar/GVN.cpp
```diff
@@ -1833,6 +1833,18 @@ bool GVN::processLoad(LoadInst *L) {
...
+      return false;
+    if (StoreInst *DepSI = dyn_cast<StoreInst>(Dep.getInst())) {
+      if (DepSI->getOperand(0)->getType()->isVectorTy() ^ L->getType()->isVectorTy())
```
sunfishcode: This can use DepSI->getValueOperand() to avoid hard-coding a magic operand number.

juj:
After this PR applied, with this .bc file: https://dl.dropboxusercontent.com/u/40949268/dump/12_PhysicsStressTest.bc , running the command

    opt 12_PhysicsStressTest.bc -emscripten-no-exit-runtime -strip-debug -disable-verify -internalize -internalize-public-api-list=main,malloc,free,__errno_location,__cxa_can_catch,__cxa_is_pointer_type,malloc,free,malloc,malloc,free,malloc,realloc,malloc,free,emscripten_GetProcAddress,malloc,free,malloc,strstr -globaldce -globalopt -globaldce -pnacl-abi-simplify-preopt -pnacl-abi-simplify-postopt -disable-loop-vectorization -disable-slp-vectorization -vectorize-loops=false -vectorize-slp=false -vectorize-slp-aggressive=false -o 12_PhysicsStressTest_out.bc

does cause an error

    Unsupported:   %41 = bitcast <4 x float> %35 to i128

but I don't know if it is `opt` creating this bad bitcast, or if it had been generated already before. I tried to `llvm-dis` the file and search where the bitcast is located, but failed. Does LLVM track enough info that it could give file:function:line information where the bad instruction occurs?

juj:
Hmm, dropbox doesn't want to sync today, so put up the same file at http://clb.demon.fi/emcc/dump/12_PhysicsStressTest.bc

juj:
Oh, I typoed the search, nm, now I found the function to be `define linkonce_odr void @_ZN12btIDebugDraw8drawConeEffiRK11btTransformRK9btVector3(%"class.Urho3D::Serializer"* %this, float %radius, float %height, i32 %upAxis, %class.btTransform* nocapture readonly dereferenceable(64) %transform, %class.btVector3* dereferenceable(16) %color) unnamed_addr #1 comdat {`..

kripken:
And is the bad IR there in the bitcode already, or introduced by that command?

juj:
It looks like the bad IR is already there in the input. It is this function: http://pastebin.com/uq3jZvsT that look like this in the IR: http://pastebin.com/hQjRQ3Lf

In particular, it contains the lines

    %41 = bitcast <4 x float> %35 to i128
    %42 = bitcast i128 %41 to <4 x i32>

but I'm not quite sure what construct creates them?

juj:
There are also other functions, for example `define void @_ZN17btGjkPairDetector26getClosestPointsNonVirtualERKN36btDiscreteCollisionDetectorInterface17ClosestPointInputERNS0_6ResultEP12btIDebugDraw(%class.btGjkPairDetector* %this, %"struct.btDiscreteCollisionDetectorInterface::ClosestPointInput"* nocapture readonly dereferenceable(144) %input, %"class.Urho3D::Serializer"* dereferenceable(4) %output, %"class.Urho3D::Serializer"* %debugDraw) #1 {`, which has

    %112 = bitcast <4 x i32> %.pre110 to i128
    %113 = bitcast i128 %112 to <4 x float>

juj:
In general it looks like the `i128` seems to always exist as a temporary to cast `<4 x i32>` to a `<4 x float>` or vice versa. I wonder why that is? Does LLVM not have an IR instruction to bitcast directly between `<4 x i32>` and `<4 x float>` and it perhaps uses `i128` as an in-between technicality of some kind? We do know how to do such bitcasts in SIMD.js, but how would we bridge this to map to those?

kripken:
Hard to say. But now that you've narrowed it down to a function, perhaps it is possible to create a standalone testcase?

kripken:
Merging with those 2 comments of @sunfishcode 's fixed. We can fix the other issues here later. I'll also land a test on the main repo.

juj:
Tried for some three hours to reduce this down to a test case, but unfortunately with no success. Anything I try to build on my own avoids the problem. The general pattern is that there's a 3-vector which is calling to negate function, which is implemented as an integer xor and an and. This creates the float->int vector cast, and after the bitops, there is a cast back. However if I copy the relevant code to a standalone code file and build that with the same flags, I am not getting the same IR, but it is always generating either a direct `<4 x float>` to a `<4 x i32>` bit cast, or a float store to memory, followed by an int load.

The same issue does occur when building ammo.js with SSE as well. STR for that:

    git clone https://github.com/juj/ammo.js
    cd ammo.js
    git checkout sse
    python make.py

Any idea why LLVM would think to introduce an intermediate `i128` types when casting a `<4 x float>` to a `<4 x i32>`? In the toy cases I write I do see that it doesn't, but in both Urho3D and ammo.js builds of Bullet they are present.

juj:
It looks like some kind of interaction with the difference between `.cpp` -> `.bc` -> `.js` vs `.cpp` -> `.js` is involved. Because if I set compile-time (`.cpp` -> `.o`) flag to `-O0` and linker flag to `-O3`, the build works without `i128`s being introduced. If instead I use compiler flag `-O3` and linker flag `-O3`, then the build fails.

kripken:
Well yes, link `-O3` doesn't do all llvm optimizations like gvn and inlining and so forth - unless LTO is requested. So that difference is expected.

I'll take a look at that ammo STR tomorrow.

kripken:
This broke some tests - looks like stopping GVN is not always a bad thing, sometimes it removes some vector operations that we don't want.

kripken:
Well, this looks quite complex. I think the original plan might be better - to legalize this pattern. Given that it is extremely rare, it seems safer to do that, and suffer some slowdown there, but not mess with GVN.

Sound ok?

juj:
In the bullet example, each `<4 x float>` to `i128` cast has a matching `i128` to `<4 x i32>` cast right after and the temporary `i128` value is not used for anything else. Would it be feasible to scan and match these pairs of instructions together and replace them with a direct `<4 x float>` -> `<4 x i32>` cast instead? (same for `<4 x i32>` -> `<4 x float>` casts?)

That would eliminate all `i128`s in the bullet case and it sounds like a safe thing to do. Perhaps if after that there are any `i128`s that remain, those could be legalized? (If it sounds messy, just legalizing all `i128`s is ok as well)

kripken:
We might do that as an optimization for things like bullet, but I see other things (that can't be optimized this way) in other testcases.

I'm surprised LLVM emits a bitcast and then a reverse bitcast - are you sure there isn't another use?

kripken:
Overall the issue is that instcombine can fix bitcasts of `A=>B=>A`, but not `A=>B=>A'` where the difference between `A` and `A'` is that they are vectors over a different type. I added support for that when the intermediary is `i128`. I also reverted these changes to SROA and GVN, as they cause more collateral damage than we can handle. Instead I pushed code to legalize i128 bitcasts to vectors.

One limitation is that in `-O1`, we fail on `test_simd15`. But we work in lower and higher opt levels. Not sure if it's worth to get `-O1` working as there is a bunch more stuff to legalize there, so I disabled it for now in that test.

kripken:
Otherwise, everything should now work on incoming, including that bullet (although it won't validate due to float64x2).

kripken:
Bullet builds, but aside from not validating as already mentioned, it does not run properly with the default `--llvm-lto 1` setting. I have that on because it was useful for normal builds, but might be an LTO bug that hits SIMD. Turning that off, it does pass the tests, however, the webgl demo in `examples/webgl` does not look right, so something might be broken.

kripken:
Is it easy to build bullet with simd? if so if we could add testing for that in the main emscripten test suite (we already test configure and cmake builds, so this could be another option in one of them) it could help. Assuming there is indeed a correctness bug that is simd-only.

juj:
Thanks @kripken, amazing work here! I pulled in your changes and I'm testing with those. Everything looks good that the build now passes, but I'm getting something computed wrong during the physics simulation when running. I'm not sure if it's related to this PR even, looking at that now.

Building bullet with simd should be quite easily doable. There are small changes that are needed locally, since they don't provide a generic option to toggle it on/off, but they gate directly on the OS+compiler in the source code, which needed readjustment. I'll look at submitting a patch for this once I verify it's running correctly.

kripken:
Nice, looks like the fix for https://github.com/kripken/emscripten/issues/3839 fixes ammo, the webgl demo now works on sse. But slow due to float64x2 breaking validation.

kripken:
Also works with `--llvm-lto 1` as well, no longer a need to remove that.

How did you narrow down that bug btw?

juj:
By an incredible stroke of luck, really. Parallel to the SIMD.js bugfix efforts, after realizing https://github.com/bulletphysics/bullet3/issues/482 , I started looking at some of the SIMD internals just out of interest for anything that would stand out, which led to this PR https://github.com/bulletphysics/bullet3/pull/485 . I was amazed to realize that those changes actually avoided the remaining problem and the physics simulation worked correctly after that, and then it was just a matter of backtracking the individual optimizations I had done. In the end, it turned out that https://github.com/juj/bullet3/commit/ef860c06c32646ff388dd87228ff27df7dce96d0 was the one that avoided the problem. I can't even begin to think how hard it would have been to dig this one out otherwise..

kripken:
Wow, that is indeed lucky :)

Earlier I was thinking about ways to debug this and I had anticipated a lot of suffering...

PR #125, opened 2015-10-22
FlattenGlobals: Remove unused BufEnd field.

waywardmonkeys:
This field was only used in an assert message, so just calculate it
at that point.


kripken:
Nice, thanks. Please just add an `// XXX EMSCRIPTEN` comment on that assert, saying that we modified this from PNaCl. We are no longer rebased on them, but perhaps we will want to sync up at some point, a comment could make that easier.

Might also want to send them this patch if you haven't already.

waywardmonkeys:
Updated. I have no idea how to send them a patch, so no plans to do so myself.

kripken:
Merged, thanks.

@jfbastien , this commit might be nice for PNaCl too.

PR #130, opened 2016-01-08
Create atomic functions in ExpandI64 if necessary

kripken:
This is for https://github.com/kripken/emscripten/issues/4025. It does two things:

1. It removes "fetch_and_" from the names, based on what you (@juj) said in that bug. But I'm confused. Is this correct?
2. It defines the methods if they are not already linked in, which was a bug.


juj:
Hey, thanks for taking this. Looks like item 2. is the correct fix for https://github.com/kripken/emscripten/issues/4025 . Reading `atomicrmw add` here http://llvm.org/docs/LangRef.html#id212, it looks like it needs to return the old value in the memory location, my comment in that bug was not precise. I think the old `fetch_and_add` form is the correct one.

kripken:
Ok, updated with those fixes.

PR #131, opened 2016-01-12
use_pthreads_2_fallback

juj:
Add casts to return values of atomic operations. Even though not strictly needed, these help implement -s USE_PTHREADS=2 build mode for backwards compatibility with non-pthreads-capable browsers.


juj:
@kripken, @lukewagner: Just to confirm, adding excessive `|0` when not needed should not cause any performance effects in asm.js codegen, but the code will be the same, right?

lukewagner:
Correct; |0 is thrown away really early in the pipeline.  Technically, the extra |0 adds a bit more parse/validate time.

kripken:
lgtm. this might actually help with other things too, like https://github.com/kripken/emscripten/issues/4018

juj:
Oh, interesting. Looks like a good fix to have then.

PR #132, opened 2016-01-22
Fix SIMD bool vector initializers

kripken:
A few of the `asm1` simd tests produce output that fails to validate, like this in `asm1.test_simd3`
````
function _emscripten_float32x4_lessThan($__a,$__b) {
 $__a = SIMD_Float32x4_check($__a);
 $__b = SIMD_Float32x4_check($__b);
 var $0 = SIMD_Bool32x4(0,0,0,0,0,0,0,0), $1 = SIMD_Int32x4(0,0,0,0), label = 0, sp = 0;
````

Note the 8 parameters to the bool constructor. I suspect that in fully optimized code, we manage to get rid of the bool temporaries, and in completely unoptimized code we don't care about asm.js validation, so we only see this in `-O1`?

(similar stuff in `test_simd6 test_sse1 test_sse1_full`)

This is fixed by this commit. However, I'm not totally sure I get that code. In particular, why not have `<= 4` etc. instead of `< 4`, which might also fix this?

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -2882,11 +2882,13 @@ void JSWriter::printFunctionBody(const Function *F) {
           // SIMD.js has only a fixed set of SIMD types, and no arbitrary vector sizes like <float x 3> or <i8 x 7>, so
           // codegen rounds up to the smallest appropriate size where the LLVM vector fits.
           unsigned simdJsNumElements = VT->getNumElements();
-          if (simdJsNumElements < 2 && VT->getElementType()->getPrimitiveSizeInBits() > 32) simdJsNumElements = 2;
-          else if (simdJsNumElements < 4 && VT->getElementType()->getPrimitiveSizeInBits() <= 32) simdJsNumElements = 4;
-          else if (simdJsNumElements < 8 && VT->getElementType()->getPrimitiveSizeInBits() <= 16) simdJsNumElements = 8;
-          else if (simdJsNumElements < 16 && VT->getElementType()->getPrimitiveSizeInBits() <= 8) simdJsNumElements = 16;
-
+          auto sizeInBits = VT->getElementType()->getPrimitiveSizeInBits();
+          if (sizeInBits > 1) {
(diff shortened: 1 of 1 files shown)


juj:
Yeah, the `<=` looks correct.

Btw, the feature of supporting code compilation of "nonstandard" vector widths like `<i8 x 3>` is something that was added only because our SIMD.js LLVM test suite has tests for that, although for SSE based compilation and the Emscripten SIMD intrinsics, those will never occur, unless one explicitly uses the Clang specific SIMD extensions e.g. with `typedef int int32x3 __attribute__((__vector_size__(12), __may_alias__));` which I suppose is legal and generates a `<i32 x 3>`.

For boolean vector types, even with the above fix commit, the codegen is not correct, but a heuristic. For example, if there is code that compares `<i32 x 3> less_than <i32 x 3>`, the result of that operation is a `<i1 x 3>` boolean vector in LLVM IR. Also, if one compares `<i8 x 3> less_than <i8 x 3>`, the result will be a `<i1 x 3>`. However, in SIMD.js, in the first case, the code would become a `Int32x4.lessThan(anotherInt32x4)` which returns a `Bool32x4`, but in the second case, it should be a `Int8x16.lessThan(anotherInt8x16)`, which returns a `Bool8x16`. This means that just looking at a type `<i1 x 3>` in LLVM IR does not precisely specify which SIMD.js boolean type should be used for that boolean, but to know exactly, the codegen should track the chain of assignments leading to the boolean until the operation is found that defines the needed SIMD.js type. This is too complex that I didn't dare to try to implement it, and the result is that only comparisons on the following LLVM SIMD types are supported: `<i32 x 1>`, `<i64 x 2>`, `<i32 x 3>`, `<i32 x 4>`, `<i16 x 5>`, `<i16 x 6>`, `<i16 x 7>`, `<i16 x 8>`, `<i8 x 9>`, `<i8 x 10>`, `<i8 x 11>`, `<i8 x 12>`, `<i8 x 13>`, `<i8 x 14>`, `<i8 x 15>` and `<i8 x 16>`.

kripken:
I see.

Ok, looks like we can close this as it was fixed by your push.

PR #133, opened 2016-01-29
simd_js_intrinsics_update

juj:
Update SIMD.js intrinsic call handlers to match the latest SIMD.js specification.


kripken:
lgtm - not that I can read every single line here ;) But I skimmed and it looks very good. Very comprehensive.

juj:
Thanks!

PR #134, opened 2016-01-29
Fix for recursively referenced insertvalue

Maher4Ever:
As you requested on kripken/emscripten#4031, here is my fix for recursively referenced insertvalue instructions. The fix removes cleaning up insertvalue instructions, which is not ideal. How can the clean up of these instructions be done recursively?

### DIFF
--- a/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
+++ b/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
@@ -490,18 +490,6 @@ static bool ExpandExtractValues(Function &Func) {
     }
   }
 
-  // Delete the insertvalue instructions.  These can reference each
-  // other, so we must do dropAllReferences() before doing
-  // eraseFromParent(), otherwise we will try to erase instructions
-  // that are still referenced.
-  for (Instruction *I : ToErase) {
-    I->dropAllReferences();
-  }
-
-  for (Instruction *I : ToErase) {
-    I->eraseFromParent();
-  }
-
   return Changed;
 }
 
--- a/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
+++ b/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
@@ -491,17 +491,17 @@ static bool ExpandExtractValues(Function &Func, bool Finalize) {
   }
 
   if (Finalize) {
-	  // Delete the insertvalue instructions. These can reference each
-	  // other, so we must do dropAllReferences() before doing
-	  // eraseFromParent(), otherwise we will try to erase instructions
-	  // that are still referenced.
-	  for (Instruction *I : ToErase) {
-		I->dropAllReferences();
-	  }
+      // Delete the insertvalue instructions. These can reference each
+      // other, so we must do dropAllReferences() before doing
+      // eraseFromParent(), otherwise we will try to erase instructions
+      // that are still referenced.
+      for (Instruction *I : ToErase) {
+        I->dropAllReferences();
+      }
 
-	  for (Instruction *I : ToErase) {
-		  I->eraseFromParent();
-	  }
+      for (Instruction *I : ToErase) {
+          I->eraseFromParent();
+      }
   }
 
   return Changed;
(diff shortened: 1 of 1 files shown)


kripken:
What I think might work is to first do the dropping as before. Then, do an infinite loop that works until there is nothing left to do, and inside it checks to see which have no uses left, and erases those, then does another iteration is it might allow more to be cleaned up. If an iteration occurs and nothing could be cleaned up, and there are still items left, then we should abort with an error.

Maher4Ever:
I modified the code to implement what you described as follows:

```cpp
  for (Instruction *I : ToErase) {
    I->dropAllReferences();
  }

  bool KeepGoing = false;

  do {
	  bool FoundSomethingToErase = false;
          
	  for (auto I = ToErase.begin(); I != ToErase.end(); ) {
		  auto Inst = *I;

		  if (Inst->getNumUses() == 0) {
			  FoundSomethingToErase = true;
			  Inst->eraseFromParent();
			  I = ToErase.erase(I);
		  }
		  else {
			  I++;
		  }
	  }
          
	  // ToErase size after: 2
	  KeepGoing = ToErase.size() != 0;
	  assert(KeepGoing == FoundSomethingToErase);
  } while (KeepGoing);
```

then I ran it with the LLVM IR I had, but the assertion I added kept failing because after running the first iteration the container contains 2 instructions (the ones recursively referenced) that still have uses and we can't erase them. Here is the LLVM IR:

```llvm
; ModuleID = 'generated.ll'
target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-windows-msvc"

define i32 @comment_set({ i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] }* noalias nocapture %retptr, { i8*, i32 }** noalias nocapture readnone %excinfo, i8* noalias nocapture readnone %env, i8* %arg.new_value.0, i8* %arg.new_value.1, i64 %arg.new_value.2, i64 %arg.new_value.3, i8* %arg.new_value.4, i64 %arg.new_value.5.0, i64 %arg.new_value.6.0, i8* %arg._value.0, i8* nocapture readnone %arg._value.1, i64 %arg._value.2, i64 %arg._value.3, i8* nocapture readnone %arg._value.4, i64 %arg._value.5.0, i64 %arg._value.6.0) {
entry:
  %inserted.meminfo = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } undef, i8* %arg.new_value.0, 0
  %inserted.parent = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.meminfo, i8* %arg.new_value.1, 1
  %inserted.nitems = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.parent, i64 %arg.new_value.2, 2
  %inserted.itemsize = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.nitems, i64 %arg.new_value.3, 3
  %inserted.data = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.itemsize, i8* %arg.new_value.4, 4
  %.19 = insertvalue [1 x i64] undef, i64 %arg.new_value.5.0, 0
  %inserted.shape = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.data, [1 x i64] %.19, 5
  %.20 = insertvalue [1 x i64] undef, i64 %arg.new_value.6.0, 0
  %inserted.strides = insertvalue { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.shape, [1 x i64] %.20, 6

  store { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] } %inserted.strides, { i8*, i8*, i64, i64, i8*, [1 x i64], [1 x i64] }* %retptr, align 8
  ret i32 0
}
```

After the first iteration and in all subsequent iterations these 2 instructions (`%inserted.shape = ...`, and `%inserted.strides = ...`) have uses left even though the operands they are referencing (I believe they are `%.19` and `%.20`) have been erased. Something I noticed is that `dropAllReferences` sets all the operands to a null pointer. I don't think this is correct though because we still need the arrays referenced in the operands.

I tried leaving these two un-erased but that didn't help either. Perhaps because the operands they reference have been erased. Ideas?

kripken:
Yeah, this is turning out to be tricky... we need to get rid of the extra uses they have.

Interestingly, they hang around after the pass normally. But the pass is run *again* later. The second time, it expands out the remaining things.

This suggested to me that we can just expand twice, and then finalize, as in this patch:

````
diff --git a/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp b/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
index b6d13f6..7c8be6e 100644
--- a/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
+++ b/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
@@ -470,7 +470,7 @@ static bool ExpandExtractValue(ExtractValueInst *EV,
   return true;
 }
 
-static bool ExpandExtractValues(Function &Func) {
+static bool ExpandExtractValues(Function &Func, bool Finalize) {
   bool Changed = false;
 
   SmallVector<Instruction *, 10> ToErase;
@@ -490,6 +490,7 @@ static bool ExpandExtractValues(Function &Func) {
     }
   }
 
+  if (Finalize) {
     // Delete the insertvalue instructions.  These can reference each
     // other, so we must do dropAllReferences() before doing
     // eraseFromParent(), otherwise we will try to erase instructions
@@ -501,15 +502,17 @@ static bool ExpandExtractValues(Function &Func) {
     for (Instruction *I : ToErase) {
       I->eraseFromParent();
     }
+  }
 
   return Changed;
 }
 
 bool ExpandStructRegs::runOnFunction(Function &Func) {
   bool Changed = false;
-  bool NeedsAnotherPass;
   const DataLayout *DL = &Func.getParent()->getDataLayout();
 
+  auto SplitStuff = [&]() {
+    bool NeedsAnotherPass;
     do {
       NeedsAnotherPass = false;
       // Split up aggregate loads, stores and phi nodes into operations on
@@ -548,8 +551,18 @@ bool ExpandStructRegs::runOnFunction(Function &Func) {
         }
       }
     } while (NeedsAnotherPass);
+  };
+
+  SplitStuff();
+  Changed |= ExpandExtractValues(Func, false);
 
-  Changed |= ExpandExtractValues(Func);
+  if (Changed) {
+    // insertvalues that receive insertvalues may require additional splitting
+    // and expansion.
+    // TODO: do we need an arbitrary amount of such passes?
+    SplitStuff();
+    ExpandExtractValues(Func, true);
+  }
 
   return Changed;
 }
````

(diff is `diff -w` for brevity)

I'm not 100% sure it's right - it might need more passes? But I can't find a testcase that hits that, so maybe not. This also looks like it has no effect on compile time, which makes sense, as most methods don't have anything to expand, and we clean out garbage even earlier than right now (that second pass right now is much later).

So this patch might be worth experimenting with. Thoughts?

**Comment by Maher4Ever** [comment]
I tried your changes today and they fixed the issue. As you already said, more passes might be needed. I haven't had time to make a test case but I suspect the current fix will fail if 3 or more levels of recursieve referencing happen in a struct (my LLVM IR contains 2 levels). I'll try testing this theory next week on Monday and I'll get back to you.

**Comment by Maher4Ever** [comment]
After quite a busy week, I finally found some time today to create a test case to check if more than 2 levels of recursive references still work with your fix. The testcase is the following:

```llvm
; ModuleID = 'test_case.ll'
target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-windows-msvc"

%nested_struct = type { [1 x i64] }
%struct = type { [1 x i64], %nested_struct }

define i32 @comment_set(%struct* noalias nocapture %retptr, i64 %arg.1, i64 %arg.2) {
entry:
  %.1 = insertvalue [1 x i64] undef, i64 %arg.1, 0
  %inserted.first = insertvalue %struct undef, [1 x i64] %.1, 0

  %.2 = insertvalue [1 x i64] undef, i64 %arg.2, 0
  %.nested_struct = insertvalue %nested_struct undef, [1 x i64] %.2, 0
  %inserted.second = insertvalue %struct %inserted.first, %nested_struct %.nested_struct, 1

  store %struct %inserted.second, %struct* %retptr, align 8
  ret i32 0
}
```

Fortunately, the code compiled and worked flawlessly. Also, I compiled all of our current LLVM code base with the patched version of fastcomp without any problems. I'll submit your changes in a commit and you can decide weather we can merge it our not. My vote would be to merge it since all tests pass and if there still exists a case in which it does not work we can fix it then.

lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
```diff
@@ -490,16 +490,18 @@ static bool ExpandExtractValues(Function &Func) {
...
-  // eraseFromParent(), otherwise we will try to erase instructions
-  // that are still referenced.
-  for (Instruction *I : ToErase) {
-    I->dropAllReferences();
-  }
-
-  for (Instruction *I : ToErase) {
-    I->eraseFromParent();
+  if (Finalize) {
+	  // Delete the insertvalue instructions. These can reference each
```
kripken: indentations look wrong - tabs perhaps?

Maher4Ever: Good catch. The windows client of Github didn't show the difference in indentation. Tabs are my default setting for VS :smile:. It's been fixed now.

lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
```diff
@@ -490,16 +490,18 @@ static bool ExpandExtractValues(Function &Func) {
...
 
-  // Delete the insertvalue instructions.  These can reference each
-  // other, so we must do dropAllReferences() before doing
-  // eraseFromParent(), otherwise we will try to erase instructions
-  // that are still referenced.
-  for (Instruction *I : ToErase) {
-    I->dropAllReferences();
-  }
+  if (Finalize) {
+      // Delete the insertvalue instructions. These can reference each
```
kripken: indentation still looks wrong though, this is 4 spaces when it should be 2?

Maher4Ever:
Sorry it took so long to fix the indentation. I had assumed it was merged and didn't check. Hopefully this will be in the next release.

kripken:
Loos good, thanks! Before we merge, can you please open a PR to the emscripten repo that adds yourself to the AUTHORS file?

kripken:
Oh, and also in that PR, please add a test. Can add the IR from the issue that led to this (or similar), to `tests/cases`, with a `.txt` file for the expected output. Then please verify `python tests/runner.py ALL.test_cases` passes.

PR #136, opened 2016-02-04
Avoid auto-declaring in call handlers

kripken:
This avoids unnecessary missing function stubs like you noticed in https://github.com/kripken/emscripten/issues/4068#issuecomment-179486581

See comment in the commit for more details.

It does come with the risk that if we forget to do a declare in a call handler, it can fail. But I think it's still worth it to avoid unnecessary code. Thoughts?

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -2,6 +2,10 @@
 //
 // Each handler needs DEF_CALL_HANDLER and SETUP_CALL_HANDLER
 //
+// Call handlers emit the code that the call will be replaced by. If that
+// emitted code contains calls, it must add the targets to Declares,
+// which are reported as declared but not implemented symbols, so that
+// JS linking brings them in.
 
 typedef std::string (JSWriter::*CallHandler)(const Instruction*, std::string Name, int NumArgs);
 typedef std::map<std::string, CallHandler> CallHandlerMap;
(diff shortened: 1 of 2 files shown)


kripken:
cc @juj

juj:
I like this! Testing now locally to see how it behaves..

juj:
If call handlers need to declare the functions (that are implemented in C side?) that they call, should `emscripten_check_longjmp` at https://github.com/kripken/emscripten-fastcomp/blob/incoming/lib/Target/JSBackend/CallHandlers.h#L223 declare `testSetjmp` and `longjmp`?

Also, `emscripten_atomic_load_f32` (https://github.com/kripken/emscripten-fastcomp/blob/incoming/lib/Target/JSBackend/CallHandlers.h#L585) should declare `_Atomics_load_f32_emulated` and `emscripten_atomic_load_f64` should declare `emscripten_atomic_load_f64`?

kripken:
1. Almost. In practice, `check_longjmp` only appears when `longjmp` does, so it's already defined (and we test this carefully). Ditto `testSetjmp`, it's brought in with `setjmp`.
2. Possibly. We have other code to bring those in, right - otherwise this wouldn't work before the pull? But, I can add those, no harm.

kripken:
Added those 2 from point 2.

juj:
Yay for code size savings!

PR #138, opened 2016-03-16
Added DWARF style debugging information extraction.

hackcasual:
More information is available here: https://gist.github.com/hackcasual/ea77cc31c6dafdda7274

This adds a pass entirely behind a flag to collect the LLVM DI entities for the program, then emits a JSON type tree from those

The debug intrinsics are designed to compile down to no-ops (however I believe they make the generated output non asm.js compatible). They are designed to be replaced by a debugger for reporting on live variables

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -216,12 +216,16 @@ namespace {
     BlockAddressMap BlockAddresses;
     std::map<std::string, AsmConstInfo> AsmConsts; // code => { index, list of seen sigs }
     NameSet FuncRelocatableExterns; // which externals are accessed in this function; we load them once at the beginning (avoids a potential call in a heap access, and might be faster)
-    unsigned MetadataNum;
-    std::map<Metadata *, unsigned> IndexedMetadata;
-    std::map<unsigned, std::string> VtableOffsets;
-    std::ostringstream TypeDebugData;
-    std::ostringstream TypeNameMap;
-    std::ostringstream FunctionMembers;
+
+    struct {
+      // 0 is reserved for void type
+      unsigned MetadataNum = 1;
+      std::map<Metadata *, unsigned> IndexedMetadata;
+      std::map<unsigned, std::string> VtableOffsets;
+      std::ostringstream TypeDebugData;
+      std::ostringstream TypeNameMap;
+      std::ostringstream FunctionMembers;
+    } cyberDWARFData;
 
     std::string CantValidate;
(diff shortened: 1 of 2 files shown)


kripken:
Overall impressive work here. And I like that it isn't intrusive.

However, LLVM DI is something that changes rapidly. We'll likely have a bunch of extra rebasing work because of this, every time we update from LLVM. Would you be up for doing that work?

hackcasual:
Yes, I'm up for it. I'm hoping that once WebAssembly completes its move to use the WebAssembly AsmPrinter, then DWARF can be emitted directly, and then switch to either generate this format directly from DWARF, or switch the Javascript debug interface over to use binary DWARF

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -206,6 +216,12 @@ namespace {
     BlockAddressMap BlockAddresses;
     std::map<std::string, AsmConstInfo> AsmConsts; // code => { index, list of seen sigs }
     NameSet FuncRelocatableExterns; // which externals are accessed in this function; we load them once at the beginning (avoids a potential call in a heap access, and might be faster)
+    unsigned MetadataNum;
+    std::map<Metadata *, unsigned> IndexedMetadata;
+    std::map<unsigned, std::string> VtableOffsets;
+    std::ostringstream TypeDebugData;
+    std::ostringstream TypeNameMap;
+    std::ostringstream FunctionMembers;
```
kripken: Where do these streams get initialized? Where does their output go?

Another question, if they are all for CyberDWARF, perhaps they could be in a struct? Or perhaps prefixed with CD? Otherwise the names (like `FunctionMembers`) don't clearly enough imply that they are for CyberDWARF.

kripken:
Sounds good.

lib/Target/JSBackend/JSBackend.cpp
hackcasual: Moved it to an anonymous struct. Streams are initialized by their default constructor. The streams are built up in a potentially interleaved way so separate streams are needed until output is ready.

TypeDebugData is the `types` member in the output JSON, TypeNameMap is the `type_name_map` member, and FunctionMembers is the `functions` member.

kripken: I see, got it.

kripken:
Looks good. I'm ok to land this now before the other pull is done, unless there is further interaction between the two that might change?

hackcasual:
Good to go in standalone, all the changes are behind the 2 config flags

kripken:
Cool.

PR #139, opened 2016-03-24
Implement emscripten_atomic_fence().

juj:


kripken:
Is this a temporary hack, workaround, or something we were always missing?

juj:
We were always missing this. The SharedArrayBuffer specification does not have atomic fence operation (it briefly did have it, but people found it too dirty and it was dropped out). Applying any atomic operation has the same effect as a fence, so we perform a no-op operation to implement it. A thought was raised with memory location 0 that a JS AoT compiler would be able to pattern match and compile this to a faster fence operation where applicable.

kripken:
Ok, thanks.

PR #141, opened 2016-05-10
Refactor the metadata intrinsics to share code with a macro

hackcasual:
Support dumping debug intrinsics

Move intrinsic metadata function arguments to the CyberDWARF file. This will make WebAssembly integration easier

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -468,63 +468,45 @@ DEF_CALL_HANDLER(llvm_expect_i1, {
   return getAssign(CI) + getValueAsStr(CI->getOperand(0));
 })
 
-DEF_CALL_HANDLER(llvm_dbg_declare, {
+std::string IntrinsicFunctionNames[2][2] = { { "metadata_llvm_dbg_value_local_d", "metadata_llvm_dbg_value_local_i" },
+                                   { "metadata_llvm_dbg_value_constant_d", "metadata_llvm_dbg_value_constant_i" } };
+
+std::string makeMetadataIntrinsic(const Instruction *CI, std::string vo, size_t lmdi, size_t dwopi ) {
   if (!EnableCyberDWARF || !EnableCyberDWARFIntrinsics)
     return "";
 
-  auto VariableOffset = "0";
+  auto VariableOffset = vo;
   auto AssignedValue = cast<MetadataAsValue>(CI->getOperand(0))->getMetadata();
-  auto const LocalVariableMD = cast<MetadataAsValue>(CI->getOperand(1))->getMetadata();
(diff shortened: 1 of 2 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -3811,6 +3824,20 @@ std::string JSWriter::generateDebugRecordForVar(Metadata *MD) {
     cyberDWARFData.TypeDebugData << VarIDForJSON << ":"
     << "[6,\"" << E->getName().str() << "\"," << E->getValue() << "],";
   }
+  else if (DIExpression *E = dyn_cast<DIExpression>(MD)) {
+    // DIExrpressions are associated with metadata intrinsics
+    // They represent a sequence of DWARF operations or arguments to those operations
+    // The element values are well defined in the DWARF spec, and should be stable
+
+    cyberDWARFData.TypeDebugData << VarIDForJSON << ":"
+    << "[99";
```
kripken: please put this on the previous line

lib/Target/JSBackend/CallHandlers.h
```diff
@@ -468,63 +468,42 @@ DEF_CALL_HANDLER(llvm_expect_i1, {
...
-  } else if (auto const *ValAsAssign = dyn_cast<ConstantAsMetadata>(AssignedValue)) {
-    Declares.insert("metadata_llvm_dbg_value_constant");
-    return "_metadata_llvm_dbg_value_constant(\"" + getValueAsStr(ValAsAssign->getValue())
-    + "," + VarMD + ")";
-  }
-
+std::string IntrinsicFunctionNames[2][2] = { { std::string("metadata_llvm_dbg_value_local_d"), std::string("metadata_llvm_dbg_value_local_i") },
+                                   { std::string("metadata_llvm_dbg_value_constant_d"), std::string("metadata_llvm_dbg_value_constant_i") } };
+
+#define DEF_DEBUG_METADATA_INTRINSIC( vo, lmdi, dwopi ) \
```
kripken: could this be a normal method instead of a macro?

kripken:
How will this help with WebAssembly?

hackcasual:
Previously the debug intrinsics looked like this

metadata_llvm_dbg_value_local($localvar, 123, 0, 9, "localvar").

While fine JavaScript, it breaks asm.js (and WebAssembly) by not differentiating between double/int for the passed in value and using a JS string.

This change pulls the last 4 arguments into the metadata file, and emits double/int specific function calls.

kripken:
I see, thanks.

hackcasual:
I haven't tested it yet on WebAssembly, but it is a necessary prerequisite to getting it working

hackcasual:
Don't merge, this breaks MSVC builds.

PR #147, opened 2016-08-07
Fix Visual Studio 2015 build.

juj:
Visual Studio 2015 Update 3 (latest version) gets confused by

    for (A B : C)
        if (x)
        else if (y)

and it complains that the `else` statement doesn't have a matching `if`. This has something to do with the modern style of `for` statement, at least in old style `for`s, the issue doesn't occur. Adding explicit block works around the problem.

### DIFF
--- a/lib/Target/JSBackend/NaCl/NormalizeAlignment.cpp
+++ b/lib/Target/JSBackend/NaCl/NormalizeAlignment.cpp
@@ -59,8 +59,8 @@ bool NormalizeAlignment::runOnFunction(Function &F) {
   DataLayout DL(F.getParent());
   bool Modified = false;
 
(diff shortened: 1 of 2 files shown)


kripken:
Thanks!

Any other concerns, or can we merge `next-merge` to `incoming`?

juj:
After this VS2015 is happy to build, I think we should be 99% good. Currently working on fixing up the Windows bot to pass on `other.test_llvm_lit`, and it would be nice to merge the other PRs https://github.com/kripken/emscripten/pull/4467 and https://github.com/kripken/emscripten/pull/4466 before we merge, those should bring `other` suite green on the Windows bot.

PR #148, opened 2016-08-22
stack_overflow_message

juj:
Improve stack bump code to emit a specific error message for stack overflows instead of the generic one.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -1354,7 +1354,7 @@ std::string JSWriter::getStackBump(unsigned Size) {
 std::string JSWriter::getStackBump(const std::string &Size) {
   std::string ret = "STACKTOP = STACKTOP + " + Size + "|0;";
   if (EmscriptenAssertions) {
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1354,7 +1354,7 @@ std::string JSWriter::getStackBump(unsigned Size) {
 std::string JSWriter::getStackBump(const std::string &Size) {
   std::string ret = "STACKTOP = STACKTOP + " + Size + "|0;";
   if (EmscriptenAssertions) {
-    ret += " if ((STACKTOP|0) >= (STACK_MAX|0)) abort();";
+    ret += " if ((STACKTOP|0) >= (STACK_MAX|0)) abortStackOverflow(" + Size + "|0);";
```
kripken: looks like `emscripten.py` only includes that method if `STACK_OVERFLOW_CHECK`, so it won't always be there even in ASSERTIONS mode. I'm ok to add that function in ASSERTIONS, after which this change would be ok.

kripken: oh wait, i see `ASSERTIONS` implies `STACK_OVERFLOW_CHECK`

lgtm

PR #152, opened 2016-09-30
wasm i64 support

kripken:
This adds support for emitting i64 operations and values directly from the fastcomp backend. This is now possible since the binaryen wasm optimizer has made running the asm.js optimizer no longer necessary. Since we can now go straight from the backend into binaryen, we can keep i64s as i64s (by not running the ExpandI64 pass, and emitting a bunch of new `i64_*` intrinsics). The code is no longer valid asm.js, but that doesn't matter.

Of course we can only do this when we are in "wasm-only" mode, since sometimes we do expect to either run the asm.js, or we are actually running the asm.js optimizer (if either the user requested it explicitly, or an option the user requested mandates it). So this is behind a flag. But by default we should be able to pass that flag.

This started as an experiment, but turned out to be very easy to do. With wasm 0xc binary support we could benchmark this soon, should be benefits.

There will be corresponding binaryen and emscripten PRs (binaryen to parse the new `i64_*` intrinsics in `asm2wasm`, emscripten to pass the wasm-only flag).

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -586,7 +586,7 @@ DEF_CALL_HANDLER(llvm_ctlz_i64, {
     return CH___default__(CI, "i64_ctlz", 1);
   }
   Declares.insert("llvm_ctlz_i64");
-  return CH___default__(CI, "_llvm_ctlz_i64", 1);
+  return CH___default__(CI, "_llvm_ctlz_i64");
 })
 
 DEF_CALL_HANDLER(llvm_cttz_i64, {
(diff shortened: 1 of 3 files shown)


juj:
This is very nice! If one wanted to do both asm.js and wasm builds, with this enabled, does that mean that those should be done as two separate compilation steps? I.e. will `-s BINARYEN_METHOD="asmjs,native-asm"` disable this optimization?

kripken:
Yes, exactly, if the code must be valid asm.js for any reason - like asmjs is among the methods as you wrote, or if the user forces the js optimizer to be on, or if the js optimizer is needed because of a special option that user requested (like the emterpreter) - then we can't do this wasm-only optimization stuff, and won't have native i64s.

PR #156, opened 2016-10-28
no_bool1x128

juj:
Fix expansion of boolean simd vectors to expand lanes up to the correct lane size (2,4,8 or 16). This prevents i1xN vector constant creation from getting erroneously expanded up to a nonexisting Bool1x128 type.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -990,9 +990,29 @@ std::string JSWriter::getAssignIfNeeded(const Value *V) {
   return std::string();
 }
 
+int SIMDNumElements(VectorType *t) {
+  assert(t->getElementType()->getPrimitiveSizeInBits() <= 128);
+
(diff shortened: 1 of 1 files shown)


kripken:
lgtm, but why is this needed now? was it broken all along, or something else?

And if feasible it would be good to add a test in the other repo.

juj:
It was broken all along, found out only when building new PhysX 3.4 with SSE enabled.

The problem atm is that I don't have a small test case, I only saw and verified it to be fixed when building against PhysX 3.4 directly. I find it difficult to correlate out the place in C++ code that causes it to write up a small snippet that causes it, since I'm dealing with a 900MB JS file when debugging :| (I'll certainly do that though)

kripken:
Got it, thanks.

juj:
I don't have a small test case for this, but we've now verified this to fix up two very large SIMD utilizing codebases, so I'll push this in to help these developers.

PR #172, opened 2017-03-20
Add --emscripten-no-unaligned-access and --emscripten-unaligned-functions-list-file options

mittorn:


kripken:
I'd like to understand the motivation and use case here more. In particular, it's possible to achieve this by using a macro or special unaligned load/store commands in the C/C++ source, so for small amounts of code that is fine. Is this meant for a larger codebase?

mittorn:
I tried to use macro, but it does not work in this case:
https://github.com/FWGS/xash3d/blob/reconnection-fix/engine/client/console.c#L767
Yes, it is possible to fix with memcpy, but it will require much of work if project contains too many errors. We already fixed hundreds of  unaligned access errors while ported engine to android (vfp does not allow unaligned floats), but not every project is ported to cpu that does not allow unaligned access at all.  I think it is better to provide easier way to port software by forcing alignmeng checks in functions that works with unaligned data. Project may contain unmaintained and undocumented parts that works as is and debugging/rewriting such parts is not very good idea.

kripken:
Ok, makes sense.

Another question: wasm will support unaligned accesses by default anyhow. So just waiting for that is an option. Unless this is urgent for you?

Also, how important is the option to have the file list? An option to support unaligned loads everywhere, or not, would be simpler and better.

mittorn:
Disabling unaligned loads everywhere causes very low performance. It might be useful for separate object files, but not to all output binary

juj:
WebAssembly has shipped in Firefox and Chrome, so if you build a WebAssembly binary today (-s BINARYEN=1), you will have support for unaligned loads without hacks. Would that be enough?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1130,9 +1140,46 @@ static const char *heapNameToAtomicTypeName(const char *HeapName)
   return "";
 }
 
+bool IsForceUnlignedFunction(const std::string &name)
+{
+  // Return true if all loads/stores in this function should be threated as unaligned
+  
+  static bool initialized = false;
+  static std::set<std::string> unalignedSet;
```
juj: I'm not quite sure about how LLVM usually structures its code, though function local static data is error prone. I'd rather see static variables at least in the file namespace, since this data cannot get uninitialized.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -153,6 +154,15 @@ OnlyWebAssembly("emscripten-only-wasm",
                 cl::desc("Generate code that will only ever be used as WebAssembly, and is not valid JS or asm.js"),
                 cl::init(false));
 
+static cl::opt<bool>
+NoUnlignedAccess("emscripten-no-unaligned-access",
```
juj: Typo: "NoUnlignedAccess" -> "NoUnalignedAccess"?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1130,9 +1140,46 @@ static const char *heapNameToAtomicTypeName(const char *HeapName)
...
+  if (!initialized)
+  {
+    std::ifstream file;
+    std::string func;
+    file.open(UnalignedFunctionsList);
+    initialized = true;
+    if(!file)
+    {
+      if (!UnalignedFunctionsList.empty())
+        errs() << "emcc: warning: Could not read unaligned functions list\n";
```
juj: This should probably be a compilation error, failing to read compilation input files should not be shrugged past, that could create miscompiled executables and one would need a keen eye to catch this in a warnings list?

juj: I think it would be nicer to open and read this file early on when loading up input files, so that such function scope static variables were not used, and inputs could be validated immediately, and not when reaching first function that actually used this.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1130,9 +1140,46 @@ static const char *heapNameToAtomicTypeName(const char *HeapName)
   return "";
 }
 
+bool IsForceUnlignedFunction(const std::string &name)
```
juj: Typo: "IsForceUnlignedFunction" -> "IsForceUnalignedFunction"?

juj:
Btw, what is your workflow like to author such unaligned function lists?

What should happen if linking code and the unaligned function list contains a function name that is not present in the compiled code?

Also, PRs should go to incoming. Although before jumping to refactor, let's make sure this is actually worth it, perhaps WebAssembly is the best thing to target here, since it gives performance and size improvements as well?

mittorn:
Switched to incoming, but it seems to be broken. Many code converting StringRef to std::string fails to build.

mittorn:
Updated clang repo, seems to work ok now, but linking now needs almost 9 gigabytes of ram

PR #173, opened 2017-03-20
Use streams instead of std::to_string

TimNN:
`std::to_string` is not available on freebsd 10, so use streams instead.

cc rust-lang/rust#40123

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -7,6 +7,8 @@
 // which are reported as declared but not implemented symbols, so that
 // JS linking brings them in.
 
+#include "llvm/Support/ScopedPrinter.h"
+
 typedef std::string (JSWriter::*CallHandler)(const Instruction*, std::string Name, int NumArgs);
 typedef std::map<std::string, CallHandler> CallHandlerMap;
 CallHandlerMap CallHandlers;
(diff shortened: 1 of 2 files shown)


kripken:
Hmm, `to_string` is in c++11 which LLVM+clang require anyhow. Is this a specific bug in freebsd 10?

This clutters the code, if we do need this, could we make a utility method that does it (best would be to just implement `to_string` if it doesn't exist).

TimNN:
> Hmm, `to_string` is in c++11 which LLVM+clang require anyhow. Is this a specific bug in freebsd 10?

As far as I can tell, this is specific to freebsd. As I understand the situation, a small part part for c++11/c99 is not supported by freebsd (`std::complex`, maybe?), leading to `_GLIBCXX_USE_C99` being undefined, which in turn causes `std::to_string` to not be available. There is a bit more information in a [thread of the freebsd mailing list](https://lists.freebsd.org/pipermail/freebsd-toolchain/2013-February/000748.html).

> This clutters the code, if we do need this, could we make a utility method that does it (best would be to just implement to_string if it doesn't exist).

Sure, where would you like the utility method to go? And do you want to use `std::to_string`, if it is available (I don't know how to easily check that in C++) or are you fine with always using our own `to_string`?

kripken:
The utility method could go at the top of JSBackend.cpp, I guess. If it's hard to use `to_string` if it's there, I'm ok with the simpler solution.

However, please find out why this is just an issue in JSBackend.cpp and not the rest of LLVM. LLVM must be using something else, perhaps they already have such a utility?

TimNN:
> However, please find out why this is just an issue in JSBackend.cpp and not the rest of LLVM. LLVM must be using something else, perhaps they already have such a utility?

Good point! Apparently there is such a method in `llvm/Support/ScopedPrinter.h`, I've switched to that.

kripken:
Thanks! Looks good.

Can you please make a PR to the main emscripten repo, adding yourself to the AUTHORS file?

We don't have tests on github, how did you test this locally? I can do some local testing too before merging (but it looks pretty safe).

TimNN:
> We don't have tests on github, how did you test this locally? I can do some local testing too before merging (but it looks pretty safe).

I did a build of `emscripten-fastcomp` targeting `x86_64-unknown-linux` and cross compiled LLVM to `{i686,x86_64}-unknown-freebsd` with this patch as part of rust's LLVM 4.0 upgrade.

I didn't run any tests, however.

kripken:
Great, I ran some tests, everything seems good.

Issue #178, opened 2017-04-09
Support splitting Phi and Select over arrays.


eddyb:
[Currently missing](https://github.com/kripken/emscripten-fastcomp/blob/881bd352731d21c7117ad7e2ece347aacae83965/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp#L540-L550), with only `Store` and `Load` supporting arrays.
Results in "Unrecognized struct value", see https://github.com/rust-lang/rust/pull/40658#issuecomment-292768371.

kripken:
Thanks, PR for this in #179.

PR #179, opened 2017-04-09
More array splitting, handle select and phi

kripken:
For #178

### DIFF
--- a/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
+++ b/lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
@@ -319,6 +319,37 @@ static bool SplitUpArrayLoad(LoadInst *Load, const DataLayout *DL) {
   return NeedsAnotherPass;
 }
 
+static bool SplitUpArraySelect(SelectInst *Select, const DataLayout *DL) {
+  ArrayType *ATy = cast<ArrayType>(Select->getType());
+  Value *NewStruct = UndefValue::get(ATy);
+
+  bool NeedsAnotherPass = false;
+  // Create a separate select instruction for each struct field.
+  for (unsigned Index = 0; Index < ATy->getNumElements(); ++Index) {
+    SmallVector<unsigned, 1> EVIndexes;
+    EVIndexes.push_back(Index);
+    Value *TrueValue = ExtractValueInst::Create(Select->getTrueValue(),
+                                                EVIndexes, "", Select);
+    Value *FalseValue = ExtractValueInst::Create(Select->getFalseValue(),
+                                                 EVIndexes, "", Select);
+    SelectInst *NewSelect = new SelectInst(Select->getCondition(), TrueValue,
+                                           FalseValue, Select);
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
```diff
@@ -319,6 +319,81 @@ static bool SplitUpArrayLoad(LoadInst *Load, const DataLayout *DL) {
...
+      BasicBlock *IncomingBB = Phi->getIncomingBlock(PhiIndex);
+      Value *EV = CopyDebug(
+          ExtractValueInst::Create(Phi->getIncomingValue(PhiIndex), EVIndexes,
+                                   Phi->getName() + ".extract",
+                                   IncomingBB->getTerminator()),
+          Phi);
+      NewPhi->addIncoming(EV, IncomingBB);
+    }
+
+    // Reconstruct the original struct value.
```
eddyb: s/struct/array

lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
```diff
@@ -319,6 +319,81 @@ static bool SplitUpArrayLoad(LoadInst *Load, const DataLayout *DL) {
   return NeedsAnotherPass;
 }
 
+static bool SplitUpArraySelect(SelectInst *Select) {
+  ArrayType *ATy = cast<ArrayType>(Select->getType());
+  Value *NewStruct = UndefValue::get(ATy);
+
+  bool NeedsAnotherPass = false;
+  // Create a separate select instruction for each struct field.
```
eddyb: s/struct field/array element/

lib/Target/JSBackend/NaCl/ExpandStructRegs.cpp
```diff
@@ -319,6 +319,81 @@ static bool SplitUpArrayLoad(LoadInst *Load, const DataLayout *DL) {
...
+                                                EVIndexes, "", Select);
+    Value *FalseValue = ExtractValueInst::Create(Select->getFalseValue(),
+                                                 EVIndexes, "", Select);
+    SelectInst *NewSelect = SelectInst::Create(Select->getCondition(),
+                                               TrueValue, FalseValue,
+                                               Select->getName(),
+                                               Select);
+    NeedsAnotherPass = NeedsAnotherPass || DoAnotherPass(NewSelect);
+
+    // Reconstruct the struct value.
```
eddyb: s/struct/array

eddyb:
I am a bit surprised that it's not easy to handle load/store/select/phi for both structs and arrays a bit more uniformly, but I suppose C++ and LLVM don't make it easy 😞.

kripken:
Thanks, a bunch of places needed struct=>array renaming, fixed.

And yeah, it's a little sad we can't use a template for both. But we hit things like getElementType() needing an index for structs but not for arrays, so indeed C++ and LLVM make that tricky...

PR #181, opened 2017-04-27
emscripten-js-ffi opt to disable JS FFI mangling.

kanaka:
For JS/Web platform, calls to JS imports are wrapped to convert i64 to
i32 and f32 to f64. Likewise calls from JS into exports do the inverse
wrapping. However, for non-web embeddings this is probably undesirable.
This change provides an option to disable that wrapping and
use the original types for the call. This is used by emscripten with the JS_FFI option (https://github.com/kripken/emscripten/pull/5168). A related option is also proposed for binaryen (https://github.com/WebAssembly/binaryen/pull/984).

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -139,7 +139,7 @@ DEF_CALL_HANDLER(__default__, {
     if (NumArgs > 0) text += ",";
   }
   // this is an ffi call if we need casts, and it is not a special Math_ builtin or wasm-only intrinsic
-  bool FFI = NeedCasts;
+  bool FFI = LegalizeJavaScriptFFI && NeedCasts;
   if (FFI) {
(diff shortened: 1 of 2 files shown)


lib/Target/JSBackend/CallHandlers.h
```diff
@@ -139,7 +139,7 @@ DEF_CALL_HANDLER(__default__, {
     if (NumArgs > 0) text += ",";
   }
   // this is an ffi call if we need casts, and it is not a special Math_ builtin or wasm-only intrinsic
-  bool FFI = NeedCasts;
+  bool FFI = (JsFfi && NeedCasts);
```
kripken: No need for `(` and `)`.

As in other repo, perhaps `JavaScriptFFI` for the new var?

kripken: or maybe `legalizeJavaScriptFFI`

Issue #183, opened 2017-05-01
Wrong chunking in postSets emission on RELOCATABLE=1


fchiba:
If externs validation is generated on 100th (or multiple number of 100) accidentally, it will be splitted into two functions like below. By this, the address of externs became 0.

```
function runPostSets() {
 var temp = 0;
(snip)
 HEAP32[(gb + (7436) | 0) >> 2] = (HEAP32[(gb + (7436) | 0) >> 2] | 0) + (gb + (7040) | 0);

 temp = g$_global_var() | 0;
 runPostSets1();
}
function runPostSets1() {
 var temp = 0;

 HEAP32[(gb + (7440) | 0) >> 2] = (HEAP32[(gb + (7440) | 0) >> 2] | 0) + temp;

 HEAP32[(gb + (7792) | 0) >> 2] = (HEAP32[(gb + (7792) | 0) >> 2] | 0) + (gb + (7796) | 0);
```

`temp = g$_global_var() | 0;` and `HEAP32[(gb + (7440) | 0) >> 2] = (HEAP32[(gb + (7440) | 0) >> 2] | 0) + temp;` should be placed in same function.
	

Reproduction code is [here](https://gist.github.com/fchiba/94d4fac6fb654f507300b0c53b68c36c).

fchiba:
https://github.com/kripken/emscripten-fastcomp/pull/184

kripken:
Thanks, if you have time, it might be worth adding that testcase to the test suite in the emscripten repo.

fchiba:
Sent a pull request.
https://github.com/kripken/emscripten/pull/5207

kripken:
Merged that PR.

PR #184, opened 2017-05-01
Fix wrong chunking in postSets (#183)

fchiba:


kripken:
Wow, nice find and good fix :) Weird that we never ran into this bug before, it's been there for years...

PR #189, opened 2017-07-04
Support for mix of absolute and relative paths when generating sourcemaps

jlmatchlock:
See https://github.com/kripken/emscripten/issues/2970#issuecomment-311937107

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -38,6 +38,7 @@
 #include "llvm/Support/ErrorHandling.h"
 #include "llvm/IR/GetElementPtrTypeIterator.h"
 #include "llvm/Support/MathExtras.h"
+#include "llvm/Support/Path.h"
 #include "llvm/Support/ScopedPrinter.h"
 #include "llvm/Support/TargetRegistry.h"
 #include "llvm/IR/DebugInfo.h"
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -728,8 +729,16 @@ namespace {
         auto *Scope = cast_or_null<DIScope>(Loc.getScope());
         if (Scope) {
           StringRef File = Scope->getFilename();
-          if (Line > 0)
-            Code << " //@line " << utostr(Line) << " \"" << (File.size() > 0 ? File.str() : "?") << "\"";
+		  if (Line > 0)
+		  {
```
kripken: `{` should be on preceding line (see other code around here)

also, looks like you are using tabs perhaps? indentation should be 2 spaces.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -728,8 +729,16 @@ namespace {
...
+			  const StringRef Dir = Loc->getDirectory();
+			  const Twine currentFilePath(File);
+
+			  if (llvm::sys::path::is_absolute(currentFilePath))
```
kripken: please use `{`, `}` around the if and else bodies

jlmatchlock:
Just to let you know, been busy with work but I'll get around this as soon as I can.

vvuk:
Happy to help get this moving again if needed -- @jlmatchlock let me know!

jlmatchlock:
Sorry for the very long delay. I'm still busy with other tasks and haven't gotten around this yet.

I believe the only one remaining task is to write an unit test for this.
I'll need to confirm later but off the top of my head, the environment to reproduce the issue looked like this:
- a project with two libraries and one application (although one library should be enough)
- libraries have relative paths from the building directory to their cpp and include files
- application source files has absolute paths

I'll try to reproduce the issue with my current setup (which has changed a bit since 6 months ago) but if you could give some pointers regarding how to setup the test case this could expedite the merge (I don't have much experience in python)

Thank you for the reminder.

PR #196, opened 2017-08-23
Em asm calltype

juj:
This expands `EM_ASM` handling to allow specifying how the JS code is run in multithreaded environment: a) locally (synchronously) on the calling thread, b) synchronously proxied to main browser thread, or c) asynchronously proxied to main browser thread.

This needs a matching Emscripten side PR to land, I'll get to that at some point.

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -712,15 +712,17 @@ DEF_CALL_HANDLER(llvm_copysign_f64, {
 
 // EM_ASM support
 
-std::string handleAsmConst(const Instruction *CI) {
+// callType is one of "", "sync_on_main_thread_" or "async_on_main_thread_" and specifies
+// in which thread's context the EM_ASM block will be executed in.
+std::string handleAsmConst(const Instruction *CI, std::string callType = "") {
   unsigned Num = getNumArgOperands(CI);
   std::string Sig;
   Sig += getFunctionSignatureLetter(CI->getType());
   for (unsigned i = 1; i < Num; i++) {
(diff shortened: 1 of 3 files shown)


lib/Target/JSBackend/CallHandlers.h
```diff
@@ -712,15 +712,17 @@ DEF_CALL_HANDLER(llvm_copysign_f64, {
 
 // EM_ASM support
 
-std::string handleAsmConst(const Instruction *CI) {
+// callType is one of "", "sync_on_main_thread_" or "async_on_main_thread_" and specifies
```
kripken: how about an enum for the call type?

kripken:
Is there a writeup of the direction this is going? Stuff like this would need to happen in the wasm backend too, so deserves a wide discussion.

lib/Target/JSBackend/CallHandlers.h
juj: Enum sounds fine, I'll change to that.

juj:
I'm quite complete with the multithreading work now, and this PR ended up being the only change needed on the LLVM side. The remaining changes were mostly to jsifier, which processes the proxying model for JS libraries. The model is the same for JS libraries and EM_ASM blocks. I'll ping on this again when I have the Emscripten side PR cleaned up for looks.

juj:
This PR should now be good to merge. Updated this along with matching https://github.com/kripken/emscripten/pull/5544 to address review comments. CC @kripken, @jgravelle-google .

kripken:
No concerns from me.

jgravelle-google:
Looks reasonable to me

juj:
Perfect!

PR #199, opened 2017-09-29
Update to LLVM 5.0

dylanmckay:
This pull request updates the fastcomp LLVM library to LLVM 5.0.

It is a merge of the [llvm-project/llvm/branches/release_50](https://github.com/llvm-mirror/llvm/tree/release_50) branch, although I've used the official LLVM [git mirror](https://git.llvm.org/git/llvm) as the remote.

LLVM 5.0.1 uses the 5.0 branch, and thus this PR also adds a few commits from 5.0.1. Once LLVM tags 5.0.1 in November, we will be free to update emscripten with the new commits.


dylanmckay:
I believe @kripken has a commit or two on the `next-merge2` branch that I will need to cherry pick. 
I will do this once I've got the build compiling.

dylanmckay:
This now compiles successfully along with kripken/emscripten-fastcomp-clang#19, I have yet to run tests, or send a PR to the kripken/emscripten repo

dylanmckay:
I forgot that development is done on the `incoming` branch again - it looks like I actually fixed up every single change required to update `master` to 5.0, not `incoming`, causing me to reinvent the wheel many times over and fix conflicts that have already been fixed. On top of this, the branch now conflicts many times with `incoming`.

I will create a new branch off `incoming`, merge 5.0, and cherry-pick all of my LLVM 5.0 fixes.

dylanmckay:
Have pushed a new version of the branch, while also cherry-picking two relevant commits from `next-merge2`.

This now successfully compiles again.

dylanmckay:
So far as actual changes go, the thing to review the most would be the AttributeList API changes.

The API changed very drastically. Previously there would be an `AttributeSet` which would store function, return, and argument attributes. This class was very obtuse and hard to use.

Rather than simply port the bare-minimum of the API over (which would be extremely ugly), I refactored a lot of the attribute handling so that it works much cleaner with the current API. This shaves off some boilerplate and makes the code much harder to follow, but the previous attribute handling code was very dense in places, and my refactoring may accidentally subtly change behaviour.

juj:
Oh, great work with this!

For reference, I started a LLVM 4.0.1 merge some time ago, but have not been able to finish that one. The LLVM 4.0.1 merge is at:
  - https://github.com/kripken/emscripten-fastcomp/pull/191
  - https://github.com/kripken/emscripten-fastcomp-clang/pull/18
  - https://github.com/kripken/emscripten/pull/5400

Reading your comments about API changes, I think some of the changes were resolved already above, so that might help some. It would be greatly preferred to first merge to LLVM 4.0.1 and only after that merge to LLVM 5.0, because that would give a good data point for developers to bisect against if something breaks upstream. We would like to tag one Emscripten version at least with 4.0.1 before 5.0 version, so that would help catch any potential regressions.

aturon:
Just as a heads up: this is becoming an increasingly urgent blocker for Rust, which supports Emscripten and is hence stuck at an older version of LLVM. Is there anything we on the Rust side can do to help this along?

kripken:
I think we are close to [merging 4.0.1](https://github.com/kripken/emscripten-fastcomp/pull/191), after which we can focus on 5.0. The last blocker is https://github.com/kripken/emscripten-fastcomp/issues/202#issuecomment-341064384 which we just diagnosed (see last comment there).

Help would be welcome. In particular that issue just mentioned as diagnosed has something to do with LLVM function attributes; someone familiar with those might see the problem much faster than us.

juj:
I was able to resolve that last remaining issue, so we should be good with merging to LLVM 4.0.1.

aturon:
Thank you @juj!

juj:
Hey @dylanmckay, @aturon: I wonder if you would be able, or would know someone who might be able to review one odd change that was needed for the LLVM 4.0.1 merge branch, here:
 - https://github.com/kripken/emscripten-fastcomp/pull/191/commits/7cd52f00cc1e3a9d711a3dd8826bd02244f371df
 - see also this conversation: https://github.com/kripken/emscripten-fastcomp/pull/191#pullrequestreview-74136780

This somehow relates to the `AttributeSet` -> `AttributeList` refactor upstream. Without the above change, I was seeing that generated code was applying off by one codegen in the ExpandByVal pass, so I had to remove the `+ 1` part there to make it right. After that all tests would pass.

However the peculiar situation here is that there are a number of other similar kind of code sites in the repository that have same kind of code pattern, and it looks like those do not need to be changed. I am trying to understand why this specific commit above was needed, and what actually caused the off by one here after the LLVM 4.0 -> LLVM 4.0.1 merge update. Any help here would be appreciated!

juj:
Looks like I was able to figure this out, it was the `getParamAttributes()` function that had changed semantics.

christianscheuer:
What's the progress for enabling LLVM 5 support in Emscripten?

juj:
We just landed the merge to LLVM 4.0.1 release point this week. Merging to LLVM 5.0 can now proceed, although I wish that this work will take the new merged LLVM 4.0.1 release point as the base, so that there won't be double merges of commits between 4.0-4.0.1 (I think the original LLVM 5.0 merge started based on LLVM 4.0 code)

nikic:
I believe this PR can be closed, as an upgrade to LLVM 6.0.1 happened in the meantime.

PR #200, opened 2017-10-18
Cherrypick from 4.0.1: Fix cmake check for futimens on macOS

asRIA:
Fix for https://github.com/kripken/emscripten/issues/5418
Where on macOS <10.13 a flag HAVE_FUTIMENS was incorrectly set to 1

Cherry-pick of: https://github.com/llvm-mirror/llvm/commit/0e3a93628a91926fe1a0b45f576ebdc9ff5c250d

Can be substituted with updating LLVM to at least 4.0.1


kripken:
Thanks, looks good, and a 4.0.1 merge isn't imminent so it makes sense to cherry-pick this.

Can you directly cherry-pick that llvm commit, though, so we have the full comment etc.? Can add llvm as another git remote, fetch it, then the commit should exist and can be cherry-picked.

asRIA:
Done: https://github.com/kripken/emscripten-fastcomp/pull/201

PR #201, opened 2017-10-18
Re-apply "Fix cmake check for futimens when deploying to earlier macOS releases

asRIA:
This fixes an issue with the use of LLVM_PARALLEL_LINK_JOBS.

Original commit message:
macOS 10.13 added a new API (futimens). This API is only available on macOS 10.13
and later, but the cmake check we have in place only tests if the symbol is
present and ignores the availability attribute. Luckily we have new warning for
this and by making this warning an error the cmake check will return the correct
result.

See also rdar://problem/33992750.

Differential Revision: https://reviews.llvm.org/D37027

git-svn-id: https://llvm.org/svn/llvm-project/llvm/trunk@311965 91177308-0d34-0410-b5e6-96231b3b80d8

----

An explicit cherry-pick of https://github.com/llvm-mirror/llvm/commit/0e3a93628a91926fe1a0b45f576ebdc9ff5c250d as requested in: https://github.com/kripken/emscripten-fastcomp/pull/200
To fix: https://github.com/kripken/emscripten/issues/5418


asRIA:
@kripken - do you thing it can be merged?

kripken:
Great, thanks.

PR #207, opened 2017-11-10
sitofp

juj:
Fix vector SIToFP instruction to generate a proper type conversion rather than a bitcast. Fixes regression from #5760.


juj:
This is quite rarely used since SIMD.js is not stable, so not surprised we have these kind of brittle bugs resulting from small changes..

PR #208, opened 2017-11-16
64bit wasm atomics

juj:
Adds support for building code for wasm multithreading with 64-bit wasm atomics.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -1207,7 +1207,7 @@ static inline const char *getHeapName(int Bytes, int Integer)
 {
   switch (Bytes) {
     default: llvm_unreachable("Unsupported type");
-    case 8: return "HEAPF64";
+    case 8: return Integer ? "HEAP64" : "HEAPF64";
     case 4: return Integer ? "HEAP32" : "HEAPF32";
     case 2: return "HEAP16";
     case 1: return "HEAP8";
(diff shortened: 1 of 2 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1281,12 +1281,21 @@ std::string JSWriter::getLoad(const Instruction *I, const Value *P, Type *T, uns
   std::string Assign = getAssign(I);
   unsigned Bytes = DL->getTypeAllocSize(T);
   bool Aligned = Bytes <= Alignment || Alignment == 0;
-  if (OnlyWebAssembly) {
+  // If the operation is volatile, we'd like to generate an atomic operation for it to make sure it is "observed" in all cases
+  // and never optimized out, but if the operation is unaligned, that won't be possible since atomic operations can only
+  // run on aligned addresses. In such case, fall back to generating a regular operation, but issue a warning.
+  bool FallbackUnalignedVolatileOperation = OnlyWebAssembly && EnablePthreads && cast<LoadInst>(I)->isVolatile() && !Aligned;
```
kripken: good to check this.

please add a test in the emscripten repo later. it will be useful also to make sure the wasm backend does this properly as well, eventually.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -1422,23 +1433,34 @@ std::string JSWriter::getStore(const Instruction *I, const Value *P, Type *T, co
   assert(sep == ';'); // FIXME when we need that
   unsigned Bytes = DL->getTypeAllocSize(T);
   bool Aligned = Bytes <= Alignment || Alignment == 0;
+  // If the operation is volatile, we'd like to generate an atomic operation for it to make sure it is "observed" in all cases
+  // and never optimized out, but if the operation is unaligned, that won't be possible since atomic operations can only
+  // run on aligned addresses. In such case, fall back to generating a regular operation, but issue a warning.
+  bool FallbackUnalignedVolatileOperation = OnlyWebAssembly && EnablePthreads && cast<LoadInst>(I)->isVolatile() && !Aligned;
```
kripken: this appears twice, maybe a utility function makes sense?

lib/Target/JSBackend/JSBackend.cpp
juj: The already existing `test_pthread_*` tests do already test for this scenario, that's where I noticed this - in some of the patterns they have 64-bit volatile ops on 32-bit aligned data structures.

lib/Target/JSBackend/JSBackend.cpp
juj: Tried to do this, but it gets a bit clunky due to having to pull in the instruction and the alignment checks and/or the `DL`/`T` objects, and generates duplicate `Aligned` tests. If that part is left out, then we'll have a `bool FallbackUnalignedVolatileOperation(bool isVolatile, bool isAligned) { return  OnlyWebAssembly && EnablePthreads && isVolatile && !isAligned; }` which is not much of an interesting function.. I'll land this in this form, feel free to suggest something and I can look at it.

PR #143, opened 2016-05-24
Add support for llvm.sin.f(32|64) intrinsics

Maher4Ever:
This adds support for the llvm.sin.f(32|64) intrinsic to fastcomp. More details can be found in the issue kripken/emscripten#4338. The accompanying change to Emscripten is in pull request kripken/emscripten#4353

PR #210, opened 2017-12-23
Add llvm.minnum.f(32|64) and llvm.maxnum.f(32|64) intrinsics

varkor:
The corresponding change to Emscripten is this PR:
https://github.com/kripken/emscripten/pull/5978, and relates to this
issue: https://github.com/kripken/emscripten/issues/5976. I’m using
https://github.com/kripken/emscripten-fastcomp/pull/143 as a model.

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -697,14 +697,6 @@ DEF_CALL_HANDLER(llvm_ctpop_i64, {
   return CH___default__(CI, "_llvm_ctpop_i64");
 })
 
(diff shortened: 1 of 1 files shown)


kripken:
Thanks!

The `maxnum` ones seem to have already existed, on [line 700](https://github.com/varkor/emscripten-fastcomp/blob/d3096f7dc2301b08e75e43828055b3c514735380/lib/Target/JSBackend/CallHandlers.h#L700). We can remove the duplication by removing the old ones, I think, your additions are simpler.

varkor:
Ah, forgot to check for `maxnum` independently of `minnum`! Fixed that now.

PR #211, opened 2018-01-15
Implement 64-bit atomics call handlers for wasm

juj:


lib/Target/JSBackend/CallHandlers.h
```diff
@@ -169,7 +169,7 @@ DEF_CALL_HANDLER(__default__, {
                            // it should have 0 uses, but just to be safe
   } else if (!ActualRT->isVoidTy()) {
     unsigned FFI_IN = FFI ? ASM_FFI_IN : 0;
-    text = getAssignIfNeeded(CI) + "(" + getCast(text, ActualRT, ASM_NONSPECIFIC | FFI_IN) + ")";
```
kripken: might as well do `")" => ')'` as well?

juj: check

kripken:
This broke `test_atomic_cxx` in all `binaryen*` modes.

kripken:
@juj: Here's the bad line: a variable is read from that is never defined.

https://gist.github.com/kripken/c072072e3e956b9ae74d4d638b34bdc5#file-bad-asm-js-L5896

PR #212, opened 2018-01-16
Remove 2-alignment on reserved function pointers

aheejin:
According to kripken/emscripten#6068, it is not necessary anymore.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -425,7 +425,8 @@ namespace {
         Sig = "X";
       }
       FunctionTable &Table = FunctionTables[Sig];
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -425,7 +425,8 @@ namespace {
         Sig = "X";
       }
       FunctionTable &Table = FunctionTables[Sig];
-      unsigned MinSize = ReservedFunctionPointers ? 2*(ReservedFunctionPointers+1) : 1; // each reserved slot must be 2-aligned
+      unsigned MinSize =
```
kripken: This could be just `ReservedFunctionPointers + 1`, I think (since if that value is 0, the result is 1 anyhow)?

aheejin: Done

kripken:
Thanks, lgtm

PR #213, opened 2018-02-01
Fix 64-bit C++11 atomics when targeting wasm in non-multithreaded mode

juj:


kripken:
Another option is to let LLVM lower them - that would also avoid calling a function for them. On line 4508 there is
```
  if (!OnlyWebAssembly) {
    // if only wasm, then we can emit i64s, otherwise they must be lowered
    PM.add(createExpandI64Pass());
  } else if (!EnablePthreads) {
    PM.add(createLowerAtomicPass());
  }
```
I think that `else if` is suspect. We need to do that also if `!OnlyWebAssembly`. So this bug should go away if we remove the `else`, I think.

juj:
Hmm, that sounds like a good idea, but trying that out, `createLowerAtomicPass()` doesn't quite seem to be able to lower all atomics, but I see the call handlers still get called on the 64-bit LLVM intrinsics. I suppose we could do both here?

juj:
Tested `python tests/runner.py binaryen*`, `python tests/runner.py`, `python tests/runner.py browser.test_pthread*` and `python tests/runner.py other` to pass.

kripken:
Yeah, should be no harm in doing both. Odd LLVM can't remove all atomics in that pass.

lgtm

PR #215, opened 2018-02-08
don't emit clz/ctz/pop intrinsics as declared functions

kripken:
The callhandler already handles them properly. This removes a little bit of unnecessary code that we were emitting.


kripken:
Hmm, this fails on a ctlz case in `test_cases` :(

kripken:
Never mind, metadce removes this overhead anyhow in wasm.

PR #216, opened 2018-02-11
Fix case where a unary operator is applied to a negative floating point literal.

TannerRogalsky:
This would produce code that was parsed as a prefix decrement operator. A space between both minus signs is enough for it to be interpreted correctly.

There test case for this is here: https://github.com/kripken/emscripten/pull/6228


kripken:
Thanks :) Wow, I'm surprised a bug like this existed unnoticed for so long!

PR #218, opened 2018-02-16
Add support for reading emJsFuncs metadata

jgravelle-google:
asm.js and asm2wasm side of https://github.com/kripken/emscripten/pull/6220

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -244,6 +244,7 @@ namespace {
     StringMap Aliases;
     BlockAddressMap BlockAddresses;
     std::map<std::string, AsmConstInfo> AsmConsts; // code => { index, list of seen sigs }
+    std::map<std::string, std::string> EmJsFunctions; // name => code
     NameSet FuncRelocatableExterns; // which externals are accessed in this function; we load them once at the beginning (avoids a potential call in a heap access, and might be faster)
     std::vector<std::string> ExtraFunctions;
     std::set<const Function*> DeclaresNeedingTypeDeclarations; // list of declared funcs whose type we must declare asm.js-style with a usage, as they may not have another usage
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -681,6 +686,26 @@ namespace {
       return Id;
     }
 
+    void handleEmJsFunctions() {
+      for (Module::const_iterator II = TheModule->begin(), E = TheModule->end();
+           II != E; ++II) {
+        const Function* F = &*II;
+        StringRef EmJsPrefix("__em_js__");
```
kripken: maybe this could be a constant in the global scope?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -244,6 +244,7 @@ namespace {
     StringMap Aliases;
     BlockAddressMap BlockAddresses;
     std::map<std::string, AsmConstInfo> AsmConsts; // code => { index, list of seen sigs }
+    std::map<std::string, std::string> EmJsFunctions; // name => code
```
kripken: maybe `Js` is nicer as `JS`? no strong feelings either way.

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -681,6 +686,26 @@ namespace {
...
+        StringRef Name(F->getName());
+        if (!Name.startswith(EmJsPrefix)) {
+          continue;
+        }
+        std::string RealName = "_" + Name.slice(EmJsPrefix.size(), Name.size()).str();
+        const Instruction* I = &*F->begin()->begin();
+        const ReturnInst* Ret = cast<ReturnInst>(I);
+        const ConstantExpr* CE = cast<ConstantExpr>(Ret->getReturnValue());
+        const GlobalVariable* G = cast<GlobalVariable>(CE->getOperand(0));
+        const ConstantDataSequential* CDS = cast<ConstantDataSequential>(G->getInitializer());
```
kripken: is all this structure (global var, constant expr, etc.) guaranteed to exist even with opts?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -3667,6 +3693,9 @@ void JSWriter::printModuleBody() {
           continue;
         }
       }
+      if (EmJsFunctions.count(fullName) > 0) {
```
kripken: Need a comment for this

lib/Target/JSBackend/JSBackend.cpp
jgravelle-google: Could be, probably slightly nicer that way.

lib/Target/JSBackend/JSBackend.cpp
jgravelle-google: I wrote it `EmJs` because it's standing for `EM_JS`, and `EMJSFunctions` seems wrong. Essentially following `AsmConsts` above, because in my mind they're sibling features.

lib/Target/JSBackend/JSBackend.cpp
jgravelle-google: Not sure how guaranteed it is, but it is the case at O0/2/3. The function is hardcoded as always being `return getelementptr i8* ... @str ...`, and given that it looks like that at O0 I'm inclined to believe it will never get more complicated than that. (The C code is similarly always `return "code here"`, so we should always see the same structure for the IR.)

kripken:
lgtm, thanks.

PR #222, opened 2018-03-07
fix_simd_right_shift

juj:
Remove generation of old SIMD.js shiftRightLogicalByScalar and shiftRightArithmeticByScalar, instead those were replaced by shiftRightByScalar, with the SIMD type Int32x4 vs Uint32x4 denoting whether the shift will be sign-extending or not.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -1138,7 +1138,7 @@ int SIMDNumElements(VectorType *t) {
   return 0;
 }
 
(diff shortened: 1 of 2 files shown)


kripken:
lgtm. is this safe to merge as is, or will it affect tests in the emscripten repo?

juj:
This fixes SIMD behavior that we did not have tests for in Emscripten, so I think this should be good to merge before merging the Emscripten side. Rebased this to latest to verify that this still builds, after which it should be good to merge in.

kripken:
lgtm, sounds good to merge.

PR #227, opened 2018-05-11
retire RELOOPERDLL_API to fix MinGW shared build

hujiajie:
With __declspec(dllexport), only the relooper APIs will be exported in the final DLL when building with MinGW. As a result, the shared build(-DBUILD_SHARED_LIBS=ON) breaks because LLVMJSBackendCodeGen.dll is not exporting all necessary symbols. The patch simply removes this
attribute. MSVC static build is supposed to be unaffected because the relooper APIs now go in LLVMJSBackendCodeGen.lib the same way as other LLVM symbols, and MSVC shared build is just not supported by LLVM.

### DIFF
--- a/tools/opt/CMakeLists.txt
+++ b/tools/opt/CMakeLists.txt
@@ -8,9 +8,6 @@ set(LLVM_LINK_COMPONENTS
   IPO
   IRReader
   Linker
-  # @LOCALMOD-BEGIN
-  JSBackend
-  # @LOCALMOD-END
   InstCombine
   Instrumentation
   MC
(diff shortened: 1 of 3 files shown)


tools/opt/CMakeLists.txt
```diff
@@ -8,9 +8,6 @@ set(LLVM_LINK_COMPONENTS
   IPO
   IRReader
   Linker
-  # @LOCALMOD-BEGIN
-  JSBackend
```
kripken: this change looks unrelated?

hujiajie: JSBackend should already be included in ${LLVM_TARGETS_TO_BUILD}. The problem here is that the line causes a cmake error if LLVM is not configured to build the JSBackend. I'm noticing this because it blocks the MinGW build on Linux, where LLVM is first built targetting X86 only.

kripken:
I see thanks. Sounds good!

PR #230, opened 2018-08-03
fix_frem_for_precise_f32

juj:
Fix FRem handling for single precision floats so that it validates as asm.js.

For code

```c
float x = ...;
fmodf(x, 3.14f);
```

fastcomp is now generating

```
$3 = Math_fround($2 % Math_fround(3.1400001));
```

yielding

```
TypeError: asm.js type error: modulo cannot receive float arguments
```

This changes the code to emit

```
$3 = Math_fround(+($2) % +(Math_fround(3.1400001)));
```

so that it validates properly.

Unity is currently hitting this in asm.js builds, in really old code that has not changed in years. I don't know why this all of a sudden misvalidates - perhaps LLVM update changes `fmodf` to no longer go to a function call to musl? Or Firefox behavior change? (Firefox 61.0.1 (64-bit) on Windows gives the error)

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -2668,7 +2668,10 @@ void JSWriter::generateExpression(const User *I, raw_string_ostream& Code) {
       case Instruction::FAdd: Code << ensureFloat(getValueAsStr(I->getOperand(0)) + " + " + getValueAsStr(I->getOperand(1)), I->getType()); break;
       case Instruction::FMul: Code << ensureFloat(getValueAsStr(I->getOperand(0)) + " * " + getValueAsStr(I->getOperand(1)), I->getType()); break;
       case Instruction::FDiv: Code << ensureFloat(getValueAsStr(I->getOperand(0)) + " / " + getValueAsStr(I->getOperand(1)), I->getType()); break;
-      case Instruction::FRem: Code << ensureFloat(getValueAsStr(I->getOperand(0)) + " % " + getValueAsStr(I->getOperand(1)), I->getType()); break;
+      case Instruction::FRem:
+        if (PreciseF32 && !I->getType()->isDoubleTy()) Code << ensureFloat("+(" + getValueAsStr(I->getOperand(0)) + ") % +(" + getValueAsStr(I->getOperand(1)) + ")", I->getType());
+        else Code << ensureFloat(getValueAsStr(I->getOperand(0)) + " % " + getValueAsStr(I->getOperand(1)), I->getType());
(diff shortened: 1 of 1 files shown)


juj:
Test in https://github.com/kripken/emscripten/pull/6944, although I think the bots might not be testing for validation, since that would require a SpiderMonkey installation that neither Amazon bots or Travis CI bots have?

kripken:
Ah yes, `/` works in asm.js floats but not `%`. I suspect we called the fmod function before which led to this not being noticed.

I don't think we need the `ensureFloat` here - we just need to do a cast to a double on both sides, for either a double or a float.

Correct on testing - we don't have spidermonkey on the bots, so asm.js isn't checked there. (But in local tests I'd notice it, maybe others.)

juj:
> I don't think we need the ensureFloat here - we just need to do a cast to a double on both sides, for either a double or a float.

Hmm, do you mean 

```c++
Code << "+(" + getValueAsStr(I->getOperand(0)) + ") % +(" + getValueAsStr(I->getOperand(1)) + ")";
```

instead of 

```c++
Code << ensureFloat("+(" + getValueAsStr(I->getOperand(0)) + ") % +(" + getValueAsStr(I->getOperand(1)) + ")", I->getType());
```

? If so, then I think there would be a mismatch with the assignment, as it would generate

```javascript
$3 = +($2) % +(Math_fround(3.1400001));
```

instead of 

```javascript
$3 = Math_fround(+($2) % +(Math_fround(3.1400001)));
```

but since it is a float operation, `$3` is initialized in the beginning of the function as `
$3 = Math_fround(0)`, so the assignment `$3 = +($2) % +(Math_fround(3.1400001));` would give a type mismatch?

kripken:
Oh right, yeah, I misread that - we do need the outer `ensureFloat`. But the two lines with the `if` could be just one line, I think, as you wrote in the comment,
```
Code << ensureFloat("+(" + getValueAsStr(I->getOperand(0)) + ") % +(" + getValueAsStr(I->getOperand(1)) + ")", I->getType());
```
`ensureFloat` handles getting the proper output type, float or double as needed? lgtm with that.

juj:
If I do that as one line like that, always casting to double, that will cause an unnecessary code size increase (and possibly a micro-slowdown for runtime performance if the `+(double)` op does not end up being recognized to be a no-op) for the case when the operands are already double. I would really prefer not to always bluntly cast to double, given that the compiler code already has identified when the operands are float or double?

kripken:
The optimizer should remove an unnecessary `+()` cast, but yeah, fair point - lgtm.

PR #231, opened 2018-09-25
Add a workaround for Safari 9 ARM iOS right shift by non-immediate zero JIT bug

juj:
I do not unfortunately have a small test case, since it somehow interacts with JIT, and simple `emscripten_request_animation_frame()` based attempts to build a repro that'd get fed to a JIT after some warm-up runs ended up empty.

The original issue occurs on stb_image codebase in function

https://github.com/nothings/stb/blob/e6afb9cbae4064da8c3e69af3ff5c4629579c1d2/deprecated/stb_image.c#L1999-L2004

on the right shift on the last return line, when loading up a png image; where it is observed that `v==16222142`, a non-zero non-negative i32 number, and `bits==16`, yielding a shift right by 0, but the expression would compute `16222142 >> 0 == 0` and return 0.

Verified this patch to fix the above `stb_image` loading code. The odd thing is that the patch was supposed to have landed on Feb 15th 2016, but iOS 9.3.5 was released only much later on August 25th, but the observed symptoms fit the description of the issue in https://bugs.webkit.org/show_bug.cgi?id=151514 perfectly, so very probable that was the cause.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -165,6 +165,18 @@ EnableCyberDWARFIntrinsics("enable-debug-intrinsics",
                            cl::desc("Include debug intrinsics in generated output"),
                            cl::init(false));
 
+// Work around Safari/WebKit bug in iOS 9.3.5: https://bugs.webkit.org/show_bug.cgi?id=151514 where computing "a >> b" or "a >>> b" in JavaScript would erroneously
+// output 0 when a!=0 and b==0, after suitable JIT compiler optimizations have been applied to a function at runtime (bug does not occur in debug builds).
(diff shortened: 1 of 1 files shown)


kripken:
Odd bug...

I'm not sure this is completely enough, though, as we may emit `a >> b` in other places. It may be better to add a JS optimizer pass for this - should be about the same amount of work. (I'm probably more familiar with that code so I could do it if you want.) May be slightly simpler that way as then it's all in a single repo.

kripken:
On the other hand, that would be slower to run, and this does seem fairly simple. lgtm.

juj:
I did go through all shifts I could find in JSBackend, and looked for non-immediate shifts that could be zero, this was the only place I found. If you know of places off the top of your head, let me know. Apart from that, iPhone 4s with iOS 9.3.5 is so old target that may not be worth spending time to look too deep into this, unless someone runs into a real problem that needs debugging further.

PR #242, opened 2018-11-13
output_used_invoke_functions

juj:
Output used invoke functions from fastcomp so that only the invoke_() functions that are needed can be generated.

It looks like `emscript_wasm_backend` function already referred to a `invoke_funcs = metadata.get('invokeFuncs', [])` field, with this contents, so adding the same to fastcomp side as well.


kripken:
lgtm. I guess this is for asm.js - for wasm the metadce optimizations remove unnecessary invokes etc.?

juj:
I was not able to see metadce optimizing the invokes out (at least in brief test in `-O3 -g2 -s WASM=1`), but after this, they do go away also in wasm.

PR #243, opened 2018-11-19
Emit helper functions callable by ftCall_<sig> for non-legal signatur…

vargaz:
…es in wasm+emulated function pointers mode.

The ftCall_<sig> functions emitted in JS don't work for signatures containing i64 types since they
directly call the wasm function through the function table:
```
function ftCall_j(x) {
  return Module['wasmTable'].get(x)();
}
```
Calling this will lead to an error like:
exception thrown: TypeError: cannot pass i64 to or from JS,ftCall_j@a.out.js:2222:10

With this change, ftCall_j can be implemented as:
```
function ftCall_j(x) {
  return Module['asm']['ftCall_helper_j'](x);
}
```
Part of the fix for:
https://github.com/kripken/emscripten/issues/7399

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -253,6 +253,7 @@ namespace {
     GlobalAddressMap GlobalAddresses;
     NameSet Externals; // vars
     NameSet Declares; // funcs
+    NameSet FtCallHelpers;
     StringMap Redirects; // library function redirects actually used, needed for wrapper funcs in tables
     std::vector<std::string> PostSets;
     NameIntMap NamedGlobals; // globals that we export as metadata to JS, so it can access them by name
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -506,7 +518,54 @@ namespace {
       }
       LegalFunc += Call + ";\n}";
       ExtraFunctions.push_back(LegalFunc);
-      return LegalName;
+	  return LegalName;
```
kripken: indentation change looks wrong for this line

vargaz: Fixed.

vargaz:
Forgot to mention that .Since its an exported function, the binaryen legalizer will add an export wrapper for it with a JS compatible signature, and the JS code will end up calling that.

kripken:
@vargaz Thanks, I think I see now. But maybe the bug is that we don't legalize the `dynCall` exports we already have?

vargaz:
ftCall_j () is supposed to be able to call everything in the wasm module, not just exported functions, this is why it needs a new helper function to call.
 There are multiple ways to fix this, and this looked like the easiest approach.

vargaz:
I don't see any dynCall methods in the .wast file for the testcase in
https://github.com/kripken/emscripten/issues/7399
Maybe they were generated before, or they need some special option to be generated.

vargaz:
How can that be done ? I tried enabling the generation of dynCalls but they end up calling the
FUNCTION_TABLE_xxx table which don't seem to be generated when ENABLE_FUNCTION_POINTERS=1 is set.

kripken:
I think this is the if where we decide if to emit dynCalls or not,

https://github.com/kripken/emscripten/blob/incoming/emscripten.py#L995

So not doing something special there, and doing the normal generation of dynCalls, may fix this bug. May be a few more places in that file where `ENABLE_FUNCTION_POINTERS` shows up and is relevant, though, I'm not sure.

vargaz:
Closing this based on comments, will make a new one using dynCalls.

PR #252, opened 2019-01-28
externUses

juj:
Track extern uses of functions and variables so that Emscripten knows to only emit those that are actually used.

### DIFF
--- a/lib/Target/JSBackend/CallHandlers.h
+++ b/lib/Target/JSBackend/CallHandlers.h
@@ -48,6 +48,27 @@ bool canInvoke(const Value *V) {
 #define DEF_CALL_HANDLER(Ident, Code) \
   std::string CH_##Ident(const Instruction *CI, std::string Name, int NumArgs=-1) { Code }
 
+void RegisterMathUse(const std::string &Name) {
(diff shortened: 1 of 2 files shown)


kripken:
Is this for DCEing of the toplevel scope of asm.js? As mentioned in the other issue, I do think improving the minification of names to also DCE unused names would not be too hard, and I can help there. Or does this do additional things?

juj:
That's right, the intent here is to track which externs LLVM codegen is adding, so that later emscripten.py can emit only the needed primitives into the asm.js module. This feels cleaner and "proper", even though it is a bit verbose. Matching PR for Emscripten is at https://github.com/emscripten-core/emscripten/pull/7944.

kripken:
I see, thanks.

Starting to read this, it looks like it has a massive diff on JSBackend.cpp, that is mostly whitespace + reordering? Was clang-tidy run on it, or something like that?

juj:
Oops, sorry. Yeah, what happened is clang-format got executed on it. Visual Studio has recently added automatic integration with clang-format and when one saves a file it runs formatting. Let me re-do the PR so it doesn't contain the formatting, I see there's a ton of changes there and it'll mud this PR.

PR #256, opened 2019-04-02
Callhanders can be called with NULL, to just detect dependencies - atomics too

kripken:
(unlikely as taking their address might seem)


mdboom:
I can confirm that this WFM.  Thanks!

PR #258, opened 2019-04-19
Perform with.overflow expansion before struct reg expansion

nikic:
Make sure ExpandArithWithOverflow runs before ExpandStructRegs. While the latter has a check to avoid processing struct returns of intrinsics, it only works when directly extracting from one. In https://github.com/rust-lang/rust/pull/58623 we ran into a case where an extractvalue of an insertvalue of a with.overflow intrinsic is performed, as shown in the test case.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -4775,6 +4775,10 @@ bool JSTargetMachine::addPassesToEmitFile(PassManagerBase &PM, raw_pwrite_stream
     // function arguments and returns into something ExpandStructRegs can expand.
     PM.add(createSimplifyStructRegSignaturesPass());
 
+    // ExpandStructRegs must be run after ExpandArithWithOverflow to expand out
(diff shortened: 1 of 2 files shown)


kripken:
Thanks for the PR! I'll run the test suite locally on this, as we don't have pre-commit testing in this repo.

Note btw that we are close to migrating to the LLVM upstream backend as our default compiler, so this backend will matter less. We hope to remove it eventually, once we can do everything in the upstream backend.

nikic:
@kripken There's been some discussion recently about phasing out the asmjs-unknown-emscripten target in rust, which uses this backend (https://internals.rust-lang.org/t/should-we-downgrade-drop-the-asmjs-target/9835), but it looks like some people are still interested in it due to IE support :/

kripken:
Looks good on the test suite, merging.

For IE and other non-wasm browsers, I'm working on wasm2js improvements currently. That's our plan for getting rid of fastcomp entirely (but will take more time after we switch away from it by default).

PR #259, opened 2019-04-27
When building relocatable code lookup function addresses dynamically via `fp$foo` functions

sbc100:
This means that the dynamic linker then assigned function pointers
and both main module and side module use `fp$foo` accessors to lookup
function addresses, just like they do for global addresses.

This allows the test_dylink_function_pointer_equality test in emscripten
to pass.

See https://github.com/emscripten-core/emscripten/issues/8268.

### DIFF
--- a/lib/Target/JSBackend/JSBackend.cpp
+++ b/lib/Target/JSBackend/JSBackend.cpp
@@ -252,6 +252,7 @@ namespace {
     AlignedHeapStartMap AlignedHeapStarts, ZeroInitStarts;
     GlobalAddressMap GlobalAddresses;
     NameSet Externals; // vars
+    NameSet ExternalFuncs; // funcs
     NameSet Declares; // funcs
     StringMap Redirects; // library function redirects actually used, needed for wrapper funcs in tables
     std::vector<std::string> Relocations;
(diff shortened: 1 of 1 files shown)


lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -973,7 +979,7 @@ namespace {
 
     // special analyses
 
-    bool canReloop(const Function *F);
+    //bool canReloop(const Function *F);
```
kripken: was this intentional?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -4653,9 +4691,11 @@ void JSWriter::calculateNativizedVars(const Function *F) {
 
 // special analyses
 
+/*
```
kripken: can just remove all this I guess?

lib/Target/JSBackend/JSBackend.cpp
```diff
@@ -3960,6 +3985,19 @@ void JSWriter::printModuleBody() {
   }
   Out << "],";
 
+  Out << "\"externFunctions\": [";
```
kripken: isn't `externs` enough for the purpose here - idn't externFunctions duplicating part of it?

lib/Target/JSBackend/JSBackend.cpp
sbc100: I think this was fixing a compiler warning.  This method is commented out in the cpp file.

sbc100: I'll make a separate change for the warning fixes.

lib/Target/JSBackend/JSBackend.cpp
sbc100: externs is for external global data and we want to be able to handle the corresponding `g$` and `fp$` helpers separately.

kripken: Can't emscripten.py tell which of the `extern`s are functions and which are globals? Then it could create the helpers there? Or is this just easier to do here?

sbc100: No. IIRC emscripten.py gets all the info it has from the metadata so we need to encode this somehow.

sbc100: I could instead list them in the declares sections with the fp$ prefix so we would know which decleres we of this type.

kripken: Sounds good either way - you know this code the best at this point.

PR #268, opened 2020-04-30
Export the names of externals that are declared as weak

sbc100:
This allows emscripten to avoid reporting errors when these are missing at link time.


sbc100:
there is at least one place in libc++abi that relies and an weak undefined function being missing at link time.  So this helps us use more libc++abi code.

PR #269, opened 2020-05-01
Export named globals even in non-relocatable code

sbc100:
This allows native code to export the address of globals using
the normal EMSCRIPTEN_KEEPALIVE mechanism.


sbc100:
This change is basically just a two line change, it removes the outer condition (use ignore-whitespce in the diff)

sbc100:
Its to allow the sbrk pointer to be exported from the native side.   I think its worth it.

